PR #4, opened 2017-09-19
Download and use java cvs client to remove dependency

sxa555:
Currently the mauve build.xml has a prerequisite of a command line cvs client on the machine it's run on in order to grab the mauve source. This PR will modify the system to pull down a java cvs client into the prereqs directory and use that to download the mauve source code, reducing the effort required to run the tests if a CVS client is not already on your machine.

(Worth pointing out that I haven't trapped error conditions, although that should be the subject of another issue as there are multiple places where failed downloads wouldn't be caught properly)


sxa555:
For reference, this allows the "three-line tactical fix" to be removed from all the stf jobs in jenkins. Retaining that code here for reference :-)

```
# Next 3 lines = tactical fudge but maybe this should be the default to remove cvs prereq https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/18
[ ! -r $WORKSPACE/systemtest_prereqs/org-netbeans-lib-cvsclient.jar ] && wget -q https://netbeans.org/projects/versioncontrol/downloads/download/org-netbeans-lib-cvsclient.jar
echo java -D$CVSROOT -jar $WORKSPACE/systemtest_prereqs/org-netbeans-lib-cvsclient.jar '$*' > $WORKSPACE/systemtest_prereqs/cvs
chmod 700 $WORKSPACE/systemtest_prereqs/cvs
```

PR #14, opened 2017-10-06
Issue 5 : Integrate system test with testkitgen

Mesbah-Alam:
Added top level build file to build all systemtest projects with
testKitGen. Also, added playlist.xml for svt tests 

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0"?>
+
+<!--
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+This is a build.xml which executes the main openjdk.build build.xml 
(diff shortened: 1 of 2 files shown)


lumpfish:
The comment says "Added top level build file to build all systemtest projects with testKitGen".  But I see an openjdk.build/include/build.xml.  Is that the 'top level' file?  Generally, how does the integration with testkitgen occur?

Mesbah-Alam:
Hi Simon, 

In order to integrate openjdk-systemtest with testKitGen, we checkout openjdk-systemtest under openjdk-test as a "project". Since testKitGen requires the build.xml of any "project" to exist inside its folder, in order to build all the svt projects under openjdk-systemtest, we needed a build.xml right under openjdk-systemtest. The function of this build.xml file is to simply invoke the "real" top level svt build file (e.g. openjdk.build/build.xml). 

For lack of a better place to put this new build.xml file, we placed it under openjdk.build/include. The Jenkins build script, as part of setup, copies this build.xml file from openjdk.build/include and places it directly under the openjdk-test/openjdk-systemtest folder during a build. 

If you think it's not the right place for it to be, we can put it somewhere else too. And I should add a top level comment inside the build.xml file itself explaining what it is used for. 

@lumpfish

lumpfish:
Thanks for the clarification.  I think it would be easier to follow if we just put the build.xml into openjdk-systemtest root.
Please could you add a comment to the file along the lines of 'a build.xml which executes the main openjdk.build build.xml to allow tools which invoke a root level build.xml file to execute a build'.

Mesbah-Alam:
The new build.xml has been moved to the root of openjdk-systemtest. Comment has been added. @lumpfish

PR #24, opened 2017-10-31
Backport systemtest build debugging options to ci.adopt systemtest build

Mesbah-Alam:

### DIFF
--- a/openjdk.build/playlist.xml
+++ b/openjdk.build/playlist.xml
@@ -4,7 +4,8 @@
 		<variations>
 			<variation></variation>
 		</variations>
-		<command>cd $(OPENJDK_SYSTEMTEST_ROOT)/openjdk.build && make test.ClassloadingLoadTest  JAVA_ARGS="$(JVM_OPTIONS)"</command>
+		<command>cd $(OPENJDK_SYSTEMTEST_ROOT)/openjdk.build && make test.ClassloadingLoadTest JAVA_ARGS=$(Q)$(JVM_OPTIONS)$(Q); \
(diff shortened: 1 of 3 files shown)


openjdk.build/playlist.xml
```diff
@@ -4,7 +4,8 @@
 		<variations>
 			<variation></variation>
 		</variations>
-		<command>cd $(OPENJDK_SYSTEMTEST_ROOT)/openjdk.build && make test.ClassloadingLoadTest  JAVA_ARGS="$(JVM_OPTIONS)"</command>
+		<command>cd $(OPENJDK_SYSTEMTEST_ROOT)/openjdk.build && make test.ClassloadingLoadTest JAVA_ARGS=$(Q)$(JVM_OPTIONS)$(Q); \
```
lumpfish: ; \ is not valid in a Windows shell.  Instead you could create a macro AND_THEN (or a similar name) with a value of '&' on windows, otherwise ';'.  and then use cd $(OPENJDK_SYSTEMTEST_ROOT)/openjdk.build && make test.ClassloadingLoadTest JAVA_ARGS=$(Q)$(JVM_OPTIONS)$(Q) $(AND_THEN) $(TEST_STATUS)

lumpfish: See https://github.com/eclipse/openj9/issues/490 for the required missing macro.  We can go ahead with the merge for now because we are not yet testing Windows.
See https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/25 for the follow on issue to change the playlist to use the macro when it exists.
--



--

PR #26, opened 2017-10-31
Issue 32: Add modes to internal systemtest builds

Mesbah-Alam:


karianna:
What does this do and can the repetition be avoided?

lumpfish:
There is a lookup (modes.xml) file which can be used to turn a 'mode' (e.g. Mode100) into a string of java -X options (e.g. -Xint -Xgcpolicy:gencon).
However, this has the downside of having to look up the mode mnemonic in order to know what the test is actually running with.  It also means that if you want to run with a one-off set of options you still have to create a 'mode', which means the modes.xml file would become very bloated.
So in (offline) discussions we decided to move away from the lookup mechanism, and instead define the full options directly in the variation tag.

Not sure what you mean by 'the repetition'.  A variation is the set of options to be applied when running that particular test - so each test has its own variations. If there are no variation tags the test is simply run with no -X options at all (which may well be fine for many tests).

The playlists for all tests run under the testkitgen mechanism are actually planned to be put together in the openjdk-tests repository (along with those for the jtreg tests etc.), in which case this pull request will not be required (there'll need to be a similar request in openjdk-tests when the directory structure for the playlist files has been determined).

smlambert:
Yes, there are a few notes regarding this change:

1) We should just proceed with the plan to create a system folder in openjdk-tests that contains a build.xml and a playlist.xml file (where the build.xml file knows what repo (openjdk-systemtests) to clone and how to build the test material, and playlist.xml file indicates which tests you'd like to run (and with what options and against what Java versions and implementations).

2) My preference for the contents of the variation tags is to put the actual options rather than the less meaningful modeNumbers (that are a carry-over from how things were done in the past).  So as a review comment <variation> Mode687</variation>  should be changed to: <variation>-XX:+UseCompressedOops -Xjit -Xgcpolicy:gencon -Xaggressive</variation>  in which case, is implementation-specific, so that <test> needs to also be tagged so that its only run against openjdk+openj9 flavours of SDK (as openjdk+hotspot would fail to parse the options).

3) I haven't looked at all of the options these modes include, but some are implementation-specific, so we need to tag the tests further so they will not be invoked against an implementation that they will not run against.

I will recommend we hold off on this PR, and move this and the build.xml file into the other repo.

PR #27, opened 2017-10-31
Add copyright header to playlist.xml file for openjdk-systemtest

Mesbah-Alam:


openjdk.build/playlist.xml
```diff
@@ -1,3 +1,16 @@
+<!--
```
renfeiw: You can add this line: \<?xml version="1.0" encoding="UTF-8"?\> before copyright comment.

Mesbah-Alam: Thanks. Added.

PR #41, opened 2017-11-13
 Allow users to pass arguments to the stf perl layer

adamfarley:
If the user wants to use stf.pl options that we don't anticipate in
the makefile, this option allows them to do that. This prevents code
duplication and dependencies between the make and perl code.


lumpfish:
Ref https://github.com/AdoptOpenJDK/stf/issues/10

PR #43, opened 2017-11-14
Change JCK platform.os setting values to Windows or other

lumpfish:
Change is required to execute on AIX.
Fixes https://github.com/AdoptOpenJDK/JCK8/issues/7


openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
```diff
@@ -463,15 +462,13 @@ public void jckConfigurationForMultijvmWithNoAgent(StfCoreExtension test) throws
...
-		// "Current system" should also be chosen for the property in the JCK Configuration editor
-		// for the template .jti files.
-		testPlatform = "Current system";
+		// Set the operating system as 'Windows' for Windows and 'other' for all other operating
+		// systems.
+		// If 'other' is specified when executing on Windows, then Windows specific settings such
+		// as systemRoot are rejected, and the JCK harness assumes that DISPLAY is required for
+		// GUI tests.
+		// Otherwise 'other' is required because the JCK harness has no inherent knowledge of AIX and
+		// zOS.
```
karianna: hanging widow (one word on a line) words are a bit annoying to the eye, but no biggy.

openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
```diff
@@ -711,8 +736,19 @@ else if ( !platform.equals("win32") ) {
...
+				// On Windows set the testplatform.os to Windows and set systemRoot, but do not
+				// set the file and path separators (an error is thrown if they are set).
+				fileContent += "set jck.env.testPlatform.os \"Windows\";\n";
+			}
+			else {
+				// On other platforms set the testplatform.os to other and set the file and path separators.
+				fileContent += "set jck.env.testPlatform.os \"other\";\n";
+				fileContent += "set jck.env.testPlatform.fileSep \"/\";\n";
+				fileContent += "set jck.env.testPlatform.pathSep \":\";\n";
+			}
```
karianna: Duplicate code?

lumpfish: Which duplication are you referring to?

lumpfish: Oh I see - the review Window wasn't expanded properly.
There are three different test suites each of which has a slightly different configuration.
At the moment the code creates the configurations files separately for each test suite - that makes it easy to see what is being written for each suite, but it does lead to some duplication where the parameters are common, as you have observed.  The alternative would be to split of various parts into methods along the lines of 'write the next part' and call those methods inside each of the test suite stanzas so the reader would have to visit those methods to see what is being set.  Unfortunately the parameters are required to appear in a particular order, so I don't think all the common parameters could be added to the configuration in one method.
Swings and roundabouts - as an observer what would you prefer?

openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
lumpfish: Moved onto preceding line.  Just happened to be how big my editor window was at the time!

openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
lumpfish: I'm merging this PR as is because we need to run on AIX and that needs the changes.  We can raise another issue to try to remove the duplicate code if we think it's needed.

PR #72, opened 2018-02-08
update ant 1.10.1 to 1.10.2 version

TianyuZuo:
* ant url link is not downloadable
* update ant url link to 1.10.2
* update ant references to 1.10.2
* this should be temporary solution since this failure is blocking regular
builds

Signed-off-by: Tianyu Zuo <tianyu@ca.ibm.com>


TianyuZuo:
this is a temporary solution to quick fix this issue since it block regular builds 

related issue https://github.com/AdoptOpenJDK/stf/issues/31

PR #80, opened 2018-02-28
Accept both Oracle and IBM values for toRadians(double)[0] test in MathAPITest

Mesbah-Alam:


Mesbah-Alam:
This PR makes the change to a new test under MathAPITest to accept both Oracle and IBM values, as explained in the discussion in: https://github.com/eclipse/openj9/issues/1001#issuecomment-362041853

Mesbah-Alam:
This creates a concern about the MathAPITest in general. We are seeing failures popping up with mismatches between Oracle and IBM results in some of these Math APIs, and we are fixing them according to the decision made originally in eclipse/openj9#1001 (comment).

It is fine with me adding in these fixes where we accept both IBM and Oracle values for the case where the difference is within 1 ulp. However, we need to make some more substantial test changes if this continues to happen in future.

PR #82, opened 2018-03-05
Fix wrong test value

Mesbah-Alam:


Mesbah-Alam:
Corrected wrong test value at test : assertEquals("toRadians(double)[15])

PR #83, opened 2018-03-06
Fix test validation value for double(1)

Mesbah-Alam:
We need to update the validation value for double(1) test from 200 to 500, as the value originally was 500.


Mesbah-Alam:
We need to update the validation value for double(1) test from 200 to 500, as the value originally was 500. Confirmed with personal build: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/362/console

PR #91, opened 2018-03-27
Update makefile to tolerate both jck8b and jck10 header file locations

sxa555:
Oracle appear to have switched the "generic UNIX" header file directory name from solaris to linux. While we could get more creative than this, the simplest solution would appear to be to let the makefile search in both locations, so that's what I've implemented. Partially fixes https://github.com/AdoptOpenJDK/stf/issues/35


sxa555:
Also added another commit to copy across the JCK9 `.jti` files - this will need some additional work as just dumping the 9 ones in the place for v10 is probably optimistic - Raised https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/92 to cover doing it properly in the not-too-distant future;-)

lumpfish:
Merging this PR which provides some of the changes requires to support Java 10 testing.  I'll start work on the remaining changes.

Issue #93, opened 2018-03-27
Update in test_targets.mk for jck10


sxa555:
Unsurprisingly the execution of JCK10 is failing as we don't have a `test.jck10` target in https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/master/openjdk.test.jck/include/test_targets.mk

This will need to be resolved by someone digging in and adding a bunch of targets as have been done for `jck8b` and `jck9`.

Issue #92, opened 2018-03-27
Create suitable .jti files for running the JCK10


sxa555:
As a tactical fudge I copied the JCK9 JTI files held in this repository  into the place where they should be for JCK10. This is almost certainly not the right solution as it hasn't been modified for JCK10 but at least allows STF to start running against Java10. This issues should cover modifying the files in [openjdk.test.jck/config](https://github.com/AdoptOpenJDK/openjdk-systemtest/tree/master/openjdk.test.jck/config)'s jck10 directory so they are suitably set up for JCK10.

(I'm running JCK10 from my branch which has the `jck9` directory [duplicated into `jck10`](https://github.com/sxa555/openjdk-systemtest/tree/JCK10_updates/openjdk.test.jck/config/jck10)

PR #98, opened 2018-04-05
Jdk10 support

lumpfish:
Fixes https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/93
Fixes https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/92
Fixes https://github.com/AdoptOpenJDK/JCK10/issues/2

I've run some of the tests from each of the devtools, runtime and compiler jck10 test suites locally on a linux_x86-64 Ubuntu 16.04.  The only failure I've seen does not appear to relate to the automation.

I've also run the vm.jni and vm.jvmti tests which I think are the ones which run the native test cases - they passed.

Integrating this code should enable full runs to be done on the AdoptOpenJDK test machines

### DIFF
--- a/openjdk.test.jck/config/jck10/compiler.jti
+++ b/openjdk.test.jck/config/jck10/compiler.jti
@@ -1,16 +1,16 @@
 #JavaTest Harness Configuration Interview
-DESCRIPTION=JCK9 compiler template jti for STF automation
+DESCRIPTION=JCK10 compiler template jti for STF automation
 INTERVIEW=com.sun.jck.interview.JCKParameters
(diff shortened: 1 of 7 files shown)


sxa555:
I'm going to kick off the automation using your branch and see what happens - seems the best test of this :-)

sxa555:
Generally LGTM and the jenkins job appears to be running. I'll let it go for a while before I click the approve button if that's ok :-)

sxa555:
LGTM @lumpfish (Ref the new commit for picking up the default excludes file)

PR #105, opened 2018-05-10
Add Java 9 modularity system tests to Adopt OpenJDK

Mesbah-Alam:
This PR delivers the Java 9 modularity system tests to the set of systemtests at Adopt. 


Mesbah-Alam:
Note: We need to find a way to have only Java 9 (and up) builds to compile this project, so that no Java 8 build attempts to compile it.

smlambert:
For compilation of systemtest, is to ensure Java9 is used for compilation, is it merely a case of setting JAVA_HOME to a Java9 SDK ?

openjdk.test.modularity/build.xml
```diff
@@ -0,0 +1,363 @@
...
+															<arg value="${command_file_@{platform}_@{release}}.sh"/>
+														</exec>
+														<delete file="${command_file_@{platform}_@{release}}.sh" verbose="true"/>
+													</else>
+												</if>
+											</sequential>
+										</for>
+									</then>
+									<else>
+										<echo message="Skipping native platform @{platform} because we are running an a ${this_platform} machine" />
```
smlambert: I won't be pointing out too many typos in this PR, but "running on a" vs "running an a"

openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/ExplicitModulesTest.java
```diff
@@ -0,0 +1,271 @@
...
+		
+		//TODO: Failure of T1 on Oracle may be related to T4 failure (PR 134665) - In which case, it's a negative test
+		// and requires updates made to the use case. 
+		
+		String moduleName = "com.test";
+		String testName = "TestImpliedReadability";
+		String testClass = "com.ibm.runtimes.test." + testName;
+		String comment = "Run " + testName;
+
+		// Running com.test.TestImpliedReadaibility with junit jar and hamcrest jar as explicit modules.
```
smlambert: Cut'n'paste typo in comment TestImpliedReadability (rather than TestImpliedReadaibility), repeated several times in code below also.

openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/JlinkPluginOptionsTest.java
```diff
@@ -0,0 +1,379 @@
...
+import java.io.*;
+import org.junit.*;
+import org.junit.runner.*;
+
+/**
+ * 
+ * User story 4) 
+ * ~~~~~~~~~~~~~
+ * As a Java Product developer I can create a custom java runtime image using various 
+ * jlink plugin options. There are, in total, 13 plugin options available with the latest jdk release.
```
smlambert: Are there still 13 plugin options?

openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/JlinkTest.java
```diff
@@ -0,0 +1,373 @@
...
+		test.doValidateFileExists("Check com.hello modular jar created", helloModuleMjar.getJarFileRef());
+		
+		/*Removing the conf folder from com.test.jlink*/
+		test.doRm("Removing conf folder from com.test.jLinkDirCopy", jLinkDirCopy.childDirectory("conf"));
+
+		/* We need to pass <arch>/*.so as input to --lib in jmod. 
+		 * This is required only for non-windows platforms.*/
+		
+		/*Creating the <arch>/*.so structure*/ 
+		if ( ! test.env().getPlatform().contains("win_x86") ) {
```
smlambert: Is this test not executed/valid on Windows?

openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/UpgradeModPathTest.java
```diff
@@ -0,0 +1,405 @@
...
+						    .addModuleToModulepath("openjdk.test.modularity/bin/tests")
+						    .addRootModule("com.test")
+						    .addModuleToUpgradeModulepath("openjdk.test.modularity/bin/common-override/java.logging")
+							.addPrereqJarToClasspath(JavaProcessDefinition.JarId.JUNIT)
+						    .addPrereqJarToClasspath(JavaProcessDefinition.JarId.HAMCREST)
+							.runClass(org.junit.runner.JUnitCore.class)
+							.addArg("com.ibm.runtimes.test.TestOverrideUsingNonUpgradablePlatformModule")
+						);
+				
+				/*Note: This behavior is slightly different in OpenJDK in that,
```
smlambert: Do we need to take this into account when running this test?  is it not applicable to openj9?

openjdk.test.modularity/build.xml
Mesbah-Alam: Fixed.

openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/JlinkPluginOptionsTest.java
Mesbah-Alam: It seems to have changed. I have removed the unnecessary line altogether from the comment. Thanks for pointing out!

Mesbah-Alam:
"For compilation of systemtest, is to ensure Java9 is used for compilation, is it merely a case of setting JAVA_HOME to a Java9 SDK ?" 

Yes, if we set JAVA_HOME to a Java 9 SDK, which is what happens during the build when we pick SE90 as Java version, `java90_compiler` property will be set to JAVA_HOME/bin/javac by top.xml and it will ensure we are using Java 9 sdk for compiling the modularity tests.

Mesbah-Alam:
Note: I have re-written some parts of the openjdk.test.modularity/build.xml script and removed code that has dependency on Ant 1.10.1 or higher (e.g. usage of modularity specific tasks and properties). I have replaced them with use of `<apply executable..>` task. This way we can use the default version of Ant that all the rest of the tests are using.

Mesbah-Alam:
Tests are building and running in this personal build: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/369/tapResults/

smlambert:
Appears like some of the test failures in the sandbox build are that setup.pl fails... have you seen those failures before?

Mesbah-Alam:
Some of the setup.pl failures are related to path issues resulting from refactoring of module names. I am debugging through them.

Mesbah-Alam:
All path issues discovered in the previous personal build were resolved. 

What's blocking us now is: the current build.xml contains a target to build natives buit it doesn't work. It is a really old, cryptic version of the build script for natives that was copied over by Mamatha (I think) when she wrote the Jlink test portion of the modularity tests, from old systemtest build scripts. It needs to be replaced with what is done in `openj9.test.sharedClasses.jvmti/build.xml`, where the story has been simplified into targets like `build-natives-windows` and `build-natives-unix`. I am trying it out locally and seeing some errors that I am debugging at the moment.

Mesbah-Alam:
The error that I receive is `error: conflicting types for 'jboolean'`
This happens due to the fact that in the makefile for jlink tests we are setting flags for the include directory from both $JAVA_HOME as well as /usr/include

` -I$(JAVA_HOME)/../include  -I$(JAVA_HOME)/include -I/usr/include` 

The problem remains if we remove /usr/include, but the problem is resolved if we keep /usr/include but remove {JAVA_HOME}/include. 

I will go with the above fix for now (as we are falling behind with open-sourcing modularity tests, but we may want to investigate why we have a conflict with ${JAVA_HOME}/include/jni.h and /usr/home/jni.h when JAVA_HOME is pointing to OpenJDK9. I will open a separate issue for this.

Mesbah-Alam:
The native test compilation issues have been resolved. All tests are being compiled and running now. We have some failures on OpenJDK9 though, as these tests were written based on IBM SDK. I will run them on OpenJDK9 OpenJ9 and compare the results. Here's the OpenJDK9 personal build run: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/11/tapResults/

Mesbah-Alam:
Updates made in tests 
~~~~~~~~~~~~~~~~~~~
1) Introduced missing tests in playlist: 
CpMpModularJarTest
CpMpModularJarTest2
CpMpModularJarTest3
CpMpTest2
CpMpTest3
JDKInternalAPIsTest
JlinkPluginOptionsTest (Updated test and removed tests for missing jlink plugins , e.g. --copy-files)
2) Test fixed: ExplicitModulesTest - Updated test for usecase1.1 to rightly fail on both OpenJ9 and OpenJDK for non-standard modules
3) Test fixed: Jlinktest - Fixed validation logic by making test look for expected failure in stdout instead of stderr -- product changed. 
4) PatchModuleTest - Fixed validation logic to look for expected exception in stderr and stdout both as the error is printed in stdout on OpenJDK and stderr on OpenJ9. 
5) PatchModuleImageTest- Fixed validation logic to look for expected exception in stderr and stdout both as the error is printed in stdout on OpenJDK and stderr on OpenJ9. 
6) UpgradeModPathTest - Fixed validation logic to look for expected exception in stderr and stdout both as the error is printed in stdout on OpenJDK and stderr on OpenJ9. 

**TODO**
We should now be left with only one failure -- ServiceLoadersTest - This test fails on both OpenJDK and OpenJ9. Need to investigate if it has become an obsolete test.

Mesbah-Alam:
Latest personal build for modularity tests: 
1) openj9: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/23/console
2) hotspot: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/22/console

openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/JlinkTest.java
Mesbah-Alam: The test should run on Windows too. I remember Mamtha stating, while she was writing this test that, on non-windows platforms, the generated native library file has to be copied over to that specific location. On Windows it is not required. Hence the copy step was done only for no-Windows platforms.

smlambert:
Agree @Mesbah-Alam - I think these tests are good to merge, and can monitor and create separate issues if any additional gotcha's percolate up.

Mesbah-Alam:
I added a check in openjdk.test.modularity/build.xml to check for JAVA_VERSION, if it is SE80, compilation is skipped, followed similar trick as used by Renfei in `openj9/test/Java9andUp/build.xml`. It works fine as seen here: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/27/console

Mesbah-Alam:
PR for playlist: https://github.com/AdoptOpenJDK/openjdk-tests/pull/425

Issue #106, opened 2018-05-25
Modularity tests break JCK builds 


Mesbah-Alam:
The addition of the new system tests for modularity have broken compilation of the JCK test jobs.

Error 
`[exec] makefile:46: *** "The variable PLATFORM needs to be defined"`

Note: 
We know currently JCKs use STF, and currently you can't get/compile STF without getting and compiling systemtests - we should open and issue and eventually fix this circular dependency.

Mesbah-Alam:
This is happening because the top.xml property `java9_platform` is not being populated properly. This could well be due to the fact that, the check inside openjdk.test.modularity/build.xml that checks for JAVA_VERSION to be above SE80 before it calls compilation of modularity test source is broken.. When it does call test compilation, it is using antcall task , which is likely not propagating the upstream properties downstream.

Mesbah-Alam:
The reason why top level properties were not being inherited by the downstream build-javah target inside openjdk.test.modularity/build.xml is because the build-javah target didn't specify dependency explicitly on the upstream target that sets those properties, i.e. `check-prereqs`. Adding that to build-javah and also build-natives targets solves the problem.

Mesbah-Alam:
The fix is delivered via : https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/108

PR #108, opened 2018-05-25
Fix Modularity test compilation issue

Mesbah-Alam:
Fix modularity test compilation issue 

### DIFF
--- a/openjdk.test.modularity/build.xml
+++ b/openjdk.test.modularity/build.xml
@@ -261,7 +261,7 @@ limitations under the License.
 		<echo message="javac-did-not-run is ${javac-did-not-run}" />
 	</target>
 	
-	<target name="build-javah" depends="check-if-javac-ran" unless="javac-did-not-run">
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/106

Mesbah-Alam:
The reason why top level properties were not being inherited by the downstream build-javah target inside openjdk.test.modularity/build.xml is because the build-javah target didn't specify dependency explicitly on the upstream target that sets those properties, i.e. check-prereqs. Adding that to build-javah and also build-natives targets solves the problem.

With this PR the problem is gone, as evident in the personal builds below: 

1) Build with OpenJDK9-OpenJ9 SDK : Here modularity tests are to be built, and they do get built without issue: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/34/console
2) Build with OpenJDK8-OpenJ9 SDK: Here modularity tests are NOT to be built, and they get skipped properly: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/35/console

PR #110, opened 2018-05-28
Update modularity test build script to use javac -h when on jdk10

Mesbah-Alam:
Update modularity test build script to use javac -h when on jdk10

### DIFF
--- a/openjdk.test.modularity/build.xml
+++ b/openjdk.test.modularity/build.xml
@@ -35,13 +35,13 @@ limitations under the License.
         
 	<property name="module" value="openjdk.test.modularity" />
     <property name="src_dir" value="src" />
-	<property name="bin_dir" value="${source_root}/${module}/./bin" /> 
+	<property name="bin_dir" value="${source_root}/${module}/bin" /> 
 	
-	<property name="native-releases" value="${JAVA_VERSION}"/>
-	<property name="jlink-testcase-src" value="${src_dir}/tests/com.test.jlink"/>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Based on http://openjdk.java.net/jeps/313, when we are using Java 10, we need to use javac with option -h when compiling the classes that require generation of header files, and the generation process will be taken care of by `javac` itself. `javah` executable has been removed from Java 10. 

So, this PR fixes the issue of systemtest compilation failure on Java 10 by updating the header file generation step in openjdk.test.modularity/build.xml to use `javah` when we are on Java 9 or Java 8, and use `javac -h` when using Java 10 or higher.

Mesbah-Alam:
Note: The unnecessary for loop to loop through native test classes has been removed, since we only have one native test that we need to compile and generate headerfile for. This reduces the complexity in the build file.

Mesbah-Alam:
- Persona build testing this PR for **SE80**: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/52/console (the test is restricted to only modularity tests, so no test was run for SE80)
- For **SE90**: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/54/consoleFull 
- For **SE100**: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/56/tapResults/ (Results of SE90 and SE100 for modularity tests are different. I'd investigate the differences in a separate issue.

Issue #111, opened 2018-05-29
openjdk-systemtest build failing due to new antcontrib prereq


lumpfish:
The openjdk-systemtest repo buil dis failing in the adtopopenjdk JCK runs with:

build:
  [taskdef] Could not load definitions from resource net/sf/antcontrib/antlib.xml. It could not be found.

build:

BUILD FAILED
/home/jenkins/workspace/JCK-fullrun-ppcle/git/openjdk-systemtest/openjdk.build/build.xml:54: The following error occurred while executing this line:
/home/jenkins/workspace/JCK-fullrun-ppcle/git/openjdk-systemtest/openjdk.test.modularity/build.xml:73: Problem: failed to create task or type if
Cause: The name is undefined.
Action: Check the spelling.
Action: Check that any custom tasks/types have been declared.
Action: Check that any <presetdef>/<macrodef> declarations have taken place.

The build was originally designed not to have a prereq on antcontrib (so that any would be used to get all the prereqs required by the build).  @Mesbah-Alam - what in the modularity tests has introduced the prereq?

Mesbah-Alam:
openjdk.test.modularity/build.xml had a <for> to traverse through a list of native tests to build. Since we have only one native test, that for loop is not essential. In fact I had removed it from the script. I will remove the ant-contrib import from the script too.

Mesbah-Alam:
Fixed via Fixed via https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/114

PR #114, opened 2018-05-29
Remove ant-contrib depdendency from openjdk.modularity.test build file

Mesbah-Alam:
Remove ant-contrib depdendency from openjdk.modularity.test build file

### DIFF
--- a/openjdk.test.modularity/build.xml
+++ b/openjdk.test.modularity/build.xml
@@ -14,7 +14,6 @@ limitations under the License.
 -->
 
 <project name="adoptopenjdk.test.modularity" default="build">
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Tested with : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/57/console

Mesbah-Alam:
Since the `for` loop has been removed from the build.xml file, we no longer have any dependency on ant-contrib. It should be removed, especially because it's causing failure in jck build, as reported in https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/111

PR #116, opened 2018-05-29
Remove <if> condition from openjdk.test.modularity/build.xml

Mesbah-Alam:
Remove <if> condition from openjdk.test.modularity/build.xml

### DIFF
--- a/openjdk.test.modularity/build.xml
+++ b/openjdk.test.modularity/build.xml
@@ -60,26 +60,20 @@ limitations under the License.
 		<path refid="junit.class.path" />	
 		<path refid="stf.class.path" />
 		<path path="${jlink-testcase-src}" />
-	</path>
-		
-		
-
+	</path>	
+	
 	<path id="module.path">
 		<path path="${bin_dir}/common-mods" />
 	</path>
 
-	<target name="build">
-		<if>
-			<equals arg1="${JAVA_VERSION}" arg2="SE80"/>
-			<then>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
The <if> task (which requires ant-contrib) is replaced with <conditon> task. This releases openjdk.test.modularity/build.xml from depending on ant-contrib.

lumpfish:
This solution depends on a property JAVA_VERSION being set - but a user may not set that.The executable being used for the compile is ${java9_compiler}.  The build already retrieves the system properties for that java, so I think there is a property ${java9.java.specification.version}.  If that contains "1.8" then java 8 is being used for the build and so the building  of the modularity tests could be skipped based on that.

Mesbah-Alam:
@lumpfish Hi Simon, 

Where is ${java9.java.specification.version} set? I could not find it in my workspace. 

JAVA_VERSION, if not set by the user, is defaulted to SE90. User may set it to SE80, SE90, SE100, valhalla etc. It is being set in the test target generating make script (testconfig). When running locally, user needs to generate the make targets via testconfig too, should take care of setting JAVA_VERSION. (most functional tests depend on JAVA_VERSION too, https://github.com/eclipse/openj9/tree/master/test/functional).

Mesbah-Alam:
Personal build testing this PR (**SE100**)- https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/59/console

**SE80**: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox3/61/console

PR #119, opened 2018-05-30
Check java version for modularity

lumpfish:


Mesbah-Alam:
```
<condition property="can_build_modularity_natives_windows" value="true">
+			<and>
+				<equals arg1="${can_build_modularity}" arg2="1.8"/>
+				<equals arg1="${is_windows}" arg2="true"/>
+			</and>
+		</condition>
+		<condition property="can_build_modularity_natives_unix" value="true">
+			<and>
+				<equals arg1="${can_build_modularity}" arg2="1.8"/>
+				<not>
+					<equals arg1="${is_windows}" arg2="true"/>
+				</not>
+			</and>
+		</condition>
```

This does not seem right. We "can" build modularity only when Java version is 9 or above. We can not build modularity when Java version is 1.8. @lumpfish

lumpfish:
Good spot! I changed the check from checking java9.java.specification.version to can_build_modularity at the last minute (thought it would read better), but forgot to change the value being checked.

Mesbah-Alam:
Manually copied over the change in this PR to my branch and tested it on Linux using personal builds (along with my change that removes "SExx" from build path of the native library).  
It works fine on:
- **SE80**: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/536/consoleFull
- **SE90**: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/544/console
- **SE100**: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/542/console

I will merge this PR now and will add my change to remove SExx from path in a separate PR afterwords.

Mesbah-Alam:
PR to remove ${JAVA_VERSION} from native lib path of jlink test: https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/121

PR #131, opened 2018-06-06
Removed ${} from referenced properties used as flags.

Mesbah-Alam:
Removed ${} from referenced properties used as flags.

### DIFF
--- a/openjdk.build/include/top.xml
+++ b/openjdk.build/include/top.xml
@@ -207,7 +207,7 @@ limitations under the License.
 		Target to get junit from https://search.maven.org/remotecontent?filepath=junit/junit/4.12/junit-4.12.jar
 		Copy to ${prereqs_root}/junit-4.12/junit-4.12.jar
 	-->
-	<target name="configure-junit-4.12" unless="${junit_4.12_available}">
+	<target name="configure-junit-4.12" unless="junit_4.12_available">
 		<download-file destdir="${prereqs_root}/junit-4.12" destfile="junit-4.12.jar" srcurl="https://search.maven.org/remotecontent?filepath=junit/junit/4.12/junit-4.12.jar"/>
 		<available file="${junit_4.12}" property="junit_4.12_available"/>
 	</target>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
When referencing conditional properties, curly braces are not required. If used, they don't resolve to their values correctly, resulting in wrong action being taken. For example, 

```
<target name="print-junit-4.12-location" if="${junit_4.12_available}">
		<echo message="Using junit-4.12 from ${junit_4.12}"/>
	</target>

	<target name="print-junit-4.12-error" unless="${junit_4.12_available}">
		<echo message="ERROR: Cannot find junit-4.12 at ${junit_4.12}"/>
		<property name="fail_run" value="true"/>
	</target>
```
Will result in: 

```
00:01:19.269 print-junit-4.12-location:
00:01:19.269 
00:01:19.269 print-junit-4.12-error:
00:01:19.269      [echo] ERROR: Cannot find junit-4.12 at /home/j9build/workspace/ibmjava8_systemtest_x86-64_linux/openjdk-tests/systemtest_prereqs/junit-4.12/junit-4.12.jar
```
(ref: https://urv-jenkins.canlab.ibm.com:9095/view/SV_Tests/job/ibmjava8_systemtest_x86-64_linux/441/consoleFull) 

However, if the curly braces are removed from `unless="${junit_4.12_available}"` and we use `unless="junit_4.12_available"` instead, correct value will be resolved ("true"), and we will not see errors such as junit not available, when it in fact is available.

Mesbah-Alam:
PR tested with: https://urv-jenkins.canlab.ibm.com:9095/view/SV_Tests/job/ibmjava8_systemtest_x86-64_linux/442/console

Mesbah-Alam:
It seems to have somehow remained hidden successfully-- I am not sure how! I discovered it today, as I was adding a new target in the same file, and didn't add curly braces for the new target's condition. It then suddenly started to expose the issue, by failing the build. It subsequently started to work again after I updated all those targets and removed the braces from the referenced properties. @sophia-guo

PR #143, opened 2018-07-05
Add -Xmx2G for OpenJ9 to trigger an expected OOM

JasonFengJ9:
Add `-Xmx2G` for `OpenJ9` to trigger an expected `OOM`

`-Xmx2G` is added for `OpenJ9` to trigger an expected `OOM` at `api/java_awt/Image_pack/WritableRaster/index.html#CreatePackedRasterTest1`, otherwise `OpenJ9` has default larger heap size in certain scenarios, no `OOM` and fails the test.

Reviewer: @smlambert 
FYI: @DanHeidinga

### DIFF
--- a/openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
+++ b/openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
@@ -257,8 +257,10 @@ public class Jck implements StfPluginInterface {
 		// If we're testing a J9 VM that will result in dumps being taken and a non-zero return code
 		// which stf will detect as a failure. So in this case add the -Xdump options required to suppress
 		// taking dumps for OutOfMemory.
+		// -Xmx2G is added for OpenJ9 to trigger an expected OOM at api/java_awt/Image_pack/WritableRaster/index.html#CreatePackedRasterTest1
+		// Otherwise OpenJ9 has default larger heap size in certain scenarios, no OOM and fails the test.
 		if (test.env().primaryJvm().isIBMJvm()) {
-			suppressOutOfMemoryDumpOptions = " -Xdump:system:none -Xdump:system:events=gpf+abort+traceassert+corruptcache -Xdump:snap:none -Xdump:snap:events=gpf+abort+traceassert+corruptcache -Xdump:java:none -Xdump:java:events=gpf+abort+traceassert+corruptcache -Xdump:heap:none -Xdump:heap:events=gpf+abort+traceassert+corruptcache";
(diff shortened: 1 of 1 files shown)


lumpfish:
This seems to be arbitrarily setting -Xmx2Gb to all tests running with an openj9 JVM.
The intent is to run the JCK with no -Xoptions (other than to suppress dumps which would be taken because some JCK tests provoke OOM conditions), thereby testing with the default jdk behaviour.
Why was this option deemed necessary - it's not obvious to me from the description?
If a machine does not have 2Gb memory wouldn't this mean the JVM may go OOM?

JasonFengJ9:
@lumpfish, The `-Xmx2G` was chosen because JCK10 test `api/java_awt/Image_pack/WritableRaster/index.html#CreatePackedRasterTest1` intends to allocate a byte array with size 2147483647 (~2G), and expects an OOM for a pass.
`OpenJ9` might set -Xmx higher than 2G by default depending on available memory hence no OOM and fail that JCK test. It was discovered that setting -Xmx2G makes the test pass.
I attempted to challenge the JCK but Oracle suggested us workaround it with the command line option discovered above.
It appears RI has default `-Xmx2G` or less, so adding `-Xmx2G` intended to match RI behaviours.

I do agree that there might be a problem if a machine running JCK tests doesn't have 2G memory which I thought is less likely considering today's machine configurations. In addition usually JCK test machines are configured specifically to run test reliably and ideally faster hence it probably is not a good idea to chose a memory constrained machine to run JCK tests.

Having said that, if it is determined that JCK tests should accommodate such machine resources, the test harness launching code need to detect current machine available memory, skip adding `-Xmx2G` if the memory is lower than 2G, and let JVM determine the default `-Xmx` value. 
Or another way is to go back to Oracle and challenge the JCK again.

JasonFengJ9:
Looked further to RI behaviours:
There are two different OOM:
one is `java.lang.OutOfMemoryError: Java heap space` just like `J9`;
the other is `java.lang.OutOfMemoryError: Requested array size exceeds VM limit`, RI appears only allow an array size `Integer.MAX_VALUE - 2`, and `Integer.MAX_VALUE - 1` (2147483646) or more will cause OOM VM limit. `J9` can create an array size up to `Integer.MAX_VALUE`. `J9` throws `java.lang.NegativeArraySizeException` when the size value is larger than `Integer.MAX_VALUE`.

This appears the actual `RI` and `J9` difference that caused original failure.
PS: RI appears have a `Xmx` default value around `2G` though.

`J9` could match `RI` behaviour to throw `java.lang.OutOfMemoryError: Requested array size exceeds VM limit` for array size `Integer.MAX_VALUE - 1`.
This PR should be reverted instead.

PR #144, opened 2018-07-12
Use jck concurrency

lumpfish:


lumpfish:
Fixes https://github.com/AdoptOpenJDK/JCK8/issues/75

sxa555:
Note, there are three windows jobs queued up at the moment and they will be subject to this change

Issue #148, opened 2018-07-24
mauve compilation errors (when trying to run Java11 tests)


smlambert:
Not sure if I am missing a prereq (doesn't appear that any setup step fails), but failing to compile when attempting to run some Java11 tests:

23:03:20     [javac] /tmp/mauve/mauve/gnu/testlet/org/omg/CORBA_2_3/ORB/Valtype/cmInfo.java:15: error: package org.omg.CORBA does not exist
23:03:20     [javac] import org.omg.CORBA.DataInputStream;
23:03:20     [javac]                     ^
23:03:20     [javac] /tmp/mauve/mauve/gnu/testlet/org/omg/CORBA_2_3/ORB/Valtype/cmInfo.java:16: error: package org.omg.CORBA does not exist
23:03:20     [javac] import org.omg.CORBA.DataOutputStream;
23:03:20     [javac]                     ^

...

23:03:21     [javac]   symbol:   class BAD_INV_ORDER
23:03:21     [javac]   location: class testAny
23:03:21     [javac] /tmp/mauve/mauve/gnu/testlet/org/omg/CORBA/Any/testAny.java:111: error: cannot find symbol
23:03:21     [javac]                             throws BAD_OPERATION
23:03:21     [javac]                                    ^

23:03:33     [javac] 100 errors
23:03:33     [javac] 46 warnings

pshipton:
CORBA is one of the packages removed from Java 11.
https://github.com/eclipse/openj9/issues/992

smlambert:
Thanks @pshipton that explains it!  

@Mesbah-Alam @lumpfish - Can someone look into this and to any other changes required in systemtest for Java11 support?  There are already some [Java11 builds ready for testing against
](https://ci.adoptopenjdk.net/view/JDK11%20WIP/)...

Mesbah-Alam:
Mauve is compiled and used as the workload in the load tests. We need to add an "EE implementation from elsewhere, such as Eclipse Jakarta", for Mauve to compile on Java 11. I will try it out locally and update this PR.

lumpfish:
[...]

Issue #150, opened 2018-07-26
Update modularity tests based on Java 11 


Mesbah-Alam:
`java.transaction` module, on which implementation of several modularity tests depend, have been completely removed in Java 11 as part of https://github.com/eclipse/openj9/issues/992

Modules removed in java 11:
   
 ```
java.xml.ws (JAX-WS, plus the related technologies SAAJ and Web Services Metadata)
    java.xml.bind (JAXB)
    java.activation (JAF)
    java.xml.ws.annotation (Common Annotations)
    java.corba (CORBA)
    java.transaction (JTA)

    java.se.ee (Aggregator module for the six modules above)
    jdk.xml.ws (Tools for JAX-WS)
    jdk.xml.bind (Tools for JAXB
```

Following tests need to be rewritten, as they depend on `java.transaction`: 
```
1) /openjdk.test.modularity/src/tests/com.test/net/adoptopenjdk/test/modularity/junit/TestStressLayers.java
2) /openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/PatchModuleImageTest.java
3) /openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/PatchModuleTest.java
4) /openjdk.test.modularity/src/com.stf/net/adoptopenjdk/test/modularity/UpgradeModPathTest.java
```

Mesbah-Alam:
Plan is to replace the role that `java.transaction` module is designed to play in modularity test with a different platform module that does exist in jdk11 , e.g. java.xml. We need to rewrite upgrade module tests, patch module tests and layers tests to validate against a method in the overridden version of java.xml.

Mesbah-Alam:
There is no "ugradable" platform module in JDK11. As suggested in : http://mail.openjdk.java.net/pipermail/jigsaw-dev/2016-September/009310.html - the only "upgradable" platform modules have been the ones related to EE -- which have been removed. Note, `java.jnlp` was upgradable, but - this has also been removed from jdk11. 

So, we should update the tests and remove the obsolete tests - e.g., tests that involve upgrading an "upgradable" platform module.

Mesbah-Alam:
Note the patch module tests still can patch "platform modules" for which we'd have to add a replacement for the deleted `java.transaction` module , e.g. `java.xml` to be used by test for patching.

Mesbah-Alam:
[...]

PR #151, opened 2018-07-26
Update systemtest to build on Java 11

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/148

### DIFF
--- a/openjdk.build/build.xml
+++ b/openjdk.build/build.xml
@@ -35,7 +35,12 @@ limitations under the License.
 	<!-- target to build all projects in the repository.  -->
 	<!-- dir must be set on the ant task otherwise the basedir property is not set to a new value in the subant task. -->
 	<!-- Also make sure stf is built -->
-	<target name="build-dependencies">
+	
+	<condition property="isJava11" value="true">
+		<equals arg1="${JAVA_VERSION}" arg2="SE110"/>
+	</condition>
+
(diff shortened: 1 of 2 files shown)


karianna:
What happens with Java 12 :-)

lumpfish:
"What happens with Java 12 :-)".
 If (when?) using -classpath becomes invalid the tests would no longer be able to run unless work was done in the mauve project to modularise the tests.  The project doesn't appear to have been active since Java 7.  Other than that tests become invalid when some backward compatibility is removed (e.g. a number of tests were excluded at Java 9 because '_' was no longer supported as an identifier).
The test suite has generally proved useful over the years for finding regressions and as a general workload for finding VM . JIT issues.

karianna:
Thanks for the explanation!  I should have been clearer in my comment.  I was referring to using a property isJava11 which possibly needs to be ifJavaGreaterThan10 or something like that and also pick up on newer versions going forwards every 6 months - might need a regex?.

Mesbah-Alam:
`isJava11` is added to temporarily exclude `test.modularity` project to be built on Java 11. This is added to unblock Java 11 test builds (jck etc) before we come up with the re-implemented modularity test cases. Once PR for https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/150 is delivered, the `isJava11` check will be removed. @karianna

Issue #155, opened 2018-07-30
Mauve compilation error on OpenJDK 11 


Mesbah-Alam:
https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/151 was previously delivered to omit mauve classes that don't compile with OpenJDK11-OpenJ9 SDK. 

However, there are a few more files that fail to compile, on OpenJDK11 - which were discovered after trying to compile systemtest on OpenJDK11 SDK, as seen in the build below: 
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/116/console

This issue will track deletion of those additional non-compliant mauve classes .

PR #156, opened 2018-07-30
Omit mauve classses that fail to compile on OpenJDK11 SDK

Mesbah-Alam:
Resolves: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/155

### DIFF
--- a/openjdk.test.mauve/build.xml
+++ b/openjdk.test.mauve/build.xml
@@ -151,7 +151,11 @@ limitations under the License.
 		<delete dir="${work_dir}"/>
 	</target>
 
-	<target name="mauve-configure" if="mauve_source_available">
+	<condition property="corbaSupported" value="true">
+		<matches string="${JAVA_VERSION}" pattern="^SE(8|9|10)0$$" />
+	</condition>
+
+	<target name="mauve-configure" if="mauve_source_available" depends="delete-corba-files">
 		<!-- Mauve ships with a gnu/testlet/config.java.in file which is processed by the
 		     configure shell script in the mauve distribution root directory to create a config.java file.
 		     The shell script cannot be run with the Windows default shell (cmd.exe) so we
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Grinder that tests this PR: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/121/console

Mesbah-Alam:
@llxia  It is expected as we still still didn't add the SE110 tags in the systemtest playlist. This test is only concerned with compiling systemtest (mauve in particular) on OpenJDK11 SDK.

openjdk.test.mauve/build.xml
```diff
@@ -327,6 +327,9 @@ limitations under the License.
 		<delete file="${src_dir}/gnu/testlet/java/util/LinkedList/AcuniaLinkedListTest.java"/>
 		<delete file="${src_dir}/gnu/testlet/java/util/Vector/AcuniaVectorTest.java"/>
 		<delete file="${src_dir}/gnu/testlet/java/lang/System/security.java"/>
+		<delete file="${src_dir}/gnu/testlet/java/lang/Thread/insecurity.java"/>
+		<delete file="${src_dir}/gnu/testlet/java/lang/Thread/security.java"/>
+		<delete file="${src_dir}/gnu/testlet/java/lang/Thread/security10.java"/>
```
llxia: What are these file used for? If we delete them for jdk11 compilation issue, they will also be deleted when testing jdk8, 9, 10. Is it intended?

Mesbah-Alam: They are part of Mauve tests which we do not run. Deleting them will not make a difference in the tests.

Mesbah-Alam: If you look in that xml file, there are many other such tests which have been excluded from compiling. We are not filtering for Java versions here at this point. This is likely something we need to look at in future.

Mesbah-Alam: @llxia - on a second thought, it doesn't hurt to add the check for Java version > 10 (from which CORBA has been removed) and only delete the CORBA related test classes if that check passes. I've added the check in this PR. Thanks for reminding!

Mesbah-Alam: Test for SE80: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/127/console

Test for SE110: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/126/console

Issue #193, opened 2018-11-20
systemtest attempts to use pre-staged mauve.jar when compiled locally with TestConfig


Mesbah-Alam:
Running systemtests locally with TestConfig will fail trying to copy mauve.jar from getDependency build. 

Steps: 
```
1. git clone https://github.com/AdoptOpenJDK/openjdk-tests.git
2.  cd /openjdk-tests
3. ./get.sh -t ../openjdk-tests -p x64_linux -i openj9 (if you are using an openj9 or ibm jdk; use 'openjdk' otherwise) 
4. cd TestConfig
5. export SPEC=linux_x86-64_cmprssptrs
6.  export BUILD_LIST=systemtest
7. Export JAVA_HOME=<JAVA_HOME_DIR>
8. Export JAVA_BIN=<JAVA_BIN_DIR>
9. make -f run_configure.mk
10. make compile
```

The compile step will end up invoking [configure_systemtest](https://github.com/AdoptOpenJDK/openjdk-tests/blob/7f959e7be0ea44844918dc1e65949aef8939764a/systemtest/build.xml#L168) which will invoke openjdk.build/build.xml with "isAutomatedBuild" flag set to true -- which will in turn run `fetch-prestaged-mauve-jar` target, instead of `run-mauve-configure` target from openjdk.build/include/top.xml. 

We have to find out a way to distinguish between whether or not we are running systemtests manually/locally using TestConfig, or in a Jenkins build environment -- and only attempt to use a pre-staged mauve.jar for the latter case. 

Note: this is not a road blocker as we can still compile and run systemtests locally using the systemtest make story.

llxia:
I did not look into the system test code for this. Just from the description, could we just download the jar if jar does not exist or sha does not match (regardless what kind build is it)? 

We do this for functional test jars https://github.com/eclipse/openj9/blob/master/test/TestConfig/scripts/tools/getDependencies.pl

Mesbah-Alam:
Yes, I am making a PR with the change.

PR #194, opened 2018-11-20
Only attempt to use pre-staged mauve jar when it is available

Mesbah-Alam:
Resolves: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/193 

### DIFF
--- a/openjdk.build/include/top.xml
+++ b/openjdk.build/include/top.xml
@@ -191,12 +191,12 @@ limitations under the License.
 	</target>
 
 	<!-- Make sure mauve.jar is built. -->
-	<target name="run-mauve-configure" depends="fetch-prestaged-mauve-jar" unless="isAutomatedBuild">
+	<target name="run-mauve-configure" depends="check-if-prestaged-prereqs-available" unless="prestaged-prereqs-available">
 		<ant antfile="${source_root}/openjdk.test.mauve/build.xml" target="configure" dir="${source_root}/openjdk.test.mauve" inheritAll="false"/>
 	</target>
 	
-	<target name="fetch-prestaged-mauve-jar" if="isAutomatedBuild">
-		<copy file="${prereqs_root}/../TestConfig/lib/systemtest_prereqs/mauve/mauve.jar" todir="${prereqs_root}/mauve"/>
+	<target name="check-if-prestaged-prereqs-available">
+		<available file="${prereqs_root}/mauve/mauve.jar" property="prestaged-prereqs-available"/>
 	</target>
 	
 	<!-- Target to check if junit 4.12 is available. -->


Mesbah-Alam:
This PR adds a target in top.xml to check if there exists any pre-staged mauve.jar to be copied over and used for running system tests. If there exists any prestaged mauve.jar, which is the case when the test is run via a Jenkins build, we simply use it. If there does not exist any prestaged mauve.jar, which is the case when we are running system tests manually, we follow the route of downloading and compiling mauve. 

The Jenkins build scenario is tested with: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/434/console

Local test was done to ensure we do invoke the targets that download and compile mauve jars when we are not in a Jenkins build.

openjdk.build/include/top.xml
```diff
@@ -191,12 +191,17 @@ limitations under the License.
 	</target>
 
 	<!-- Make sure mauve.jar is built. -->
-	<target name="run-mauve-configure" depends="fetch-prestaged-mauve-jar" unless="isAutomatedBuild">
+	
+	<target name="run-mauve-configure" unless="prestaged-prereqs-available">
```
sophia-guo: Instead of making target 'check-if-prestaged-prereqs-available' and `fetch-prestaged-mauve-jar` as a global one and scatter around I would suggest do like <target name="run-mauve-configure" depends="fetch-prestaged-mauve-jar" unless="prestaged-prereqs-available">, 

Hence <target name="fetch-prestaged-mauve-jar" depends=""check-if-prestaged-prereqs-available"  if="prestaged-prereqs-available">

sophia-guo: Or could be even more straightforward as 
`if jar is there
     cp
else 
    run mauve configure
endif
`

Mesbah-Alam: You are right about using if-else. However, the systemtest Ant scripts have been written to avoid use of if-then-else directives. I just try to keep that coding standard going :).

openjdk.build/include/top.xml
```diff
@@ -191,12 +191,17 @@ limitations under the License.
...
 	</target>
 	
-	<target name="fetch-prestaged-mauve-jar" if="isAutomatedBuild">
-		<copy file="${prereqs_root}/../TestConfig/lib/systemtest_prereqs/mauve/mauve.jar" todir="${prereqs_root}/mauve"/>
+	<target name="check-if-prestaged-prereqs-available">
+		<available file="${prereqs_root}/../TestConfig/lib/systemtest_prereqs/mauve/mauve.jar" property="prestaged-prereqs-available"/>
+	</target>
+	
+	<target name="fetch-prestaged-mauve-jar" if="prestaged-prereqs-available">
+		<copy file="${prereqs_root}/../TestConfig/lib/systemtest_prereqs/mauve/mauve.jar" todir="${prereqs_root}/mauve" failonerror="false"/>
```
llxia: Is possible to directly get the jars into `${prereqs_root}` in the first step in Jenkins file? Feel redundant to copy it into `TestConfig/lib/systemtest_prereqs`, then move it into `${prereqs_root}/mauve`.

openjdk.build/include/top.xml
Mesbah-Alam:
```Instead of making target 'check-if-prestaged-prereqs-available' and fetch-prestaged-mauve-jar as a global one and scatter around I would suggest do like
<target name="run-mauve-configure" depends="fetch-prestaged-mauve-jar" unless="prestaged-prereqs-available">``` - yes, that could be done!. I've updated the PR, please refresh.

openjdk.build/include/top.xml
Mesbah-Alam: The copyArtifacts step happens at the beginning of the build, at which point we still do not have the `{prereqs_root}/` folder - that gets created much later when we have cloned the systemtest repo and started the build process. This is why the two steps were needed. @llxia

openjdk.build/include/top.xml
Mesbah-Alam: Tested this change by locally running the system tests ; and for Jenkins scenario, in this Grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/436/console

openjdk.build/include/top.xml
Mesbah-Alam: On a second try, you are right. I can directly copy into the `{prereqs_root}` directory, instead of having a second copy step. Updated the PR. Tested here: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/25/console @llxia

PR #196, opened 2018-11-27
Update modularity tests based on Java 11 changes in SDK

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/150


Mesbah-Alam:
This PR contains the changes required to get modularity tests running on JDK11 : https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/196

Change summary 
1) Removed reference to `java.transaction` module from all tests (as this module no longer exists in JDK11) and replaced them with `java.xml`. 
2) Upgrade module path related tests have been updated to exclude tests for "upgrading an upgradale system/platform module -- as the only "upradale" platform modules were ee modules (java.transaction) which have been removed in jdk11.
3) Updated CpMpJink test and removed use of `--class-for-name` (this option is no longer supported in jdk11). 
4) Updated TestJDKInternalAPI test and removed reference to `sun.reflect.Reflection` as it no longer exists in JDK11. 
5) Removed exclusion of modularity project from system test project build list.

Mesbah-Alam:
The PR has been tested here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/465/tapResults/

Note: the two failing layers stress tests could e real issues. I'll open separate PR to investigate them, they will be excluded from the playlist.

Mesbah-Alam:
Playlist changes to include modularity tests to jdk11 is contained here: https://github.com/AdoptOpenJDK/openjdk-tests/pull/703

openjdk.build/build.xml
```diff
@@ -55,19 +51,13 @@ limitations under the License.
 		<ant antfile="${source_root}/openjdk.test.serialization/build.xml" dir="${source_root}/openjdk.test.serialization" inheritAll="false"/>
 		<ant antfile="${source_root}/openjdk.test.util/build.xml" dir="${source_root}/openjdk.test.util" inheritAll="false"/>
 		<ant antfile="${source_root}/openjdk.test.jlm/build.xml" dir="${source_root}/openjdk.test.jlm" inheritAll="false"/>
+		<ant antfile="${source_root}/openjdk.test.modularity/build.xml" dir="${source_root}/openjdk.test.modularity" inheritAll="false"/>
```
sophia-guo: Will the source be built by any jdk version or build.xml has flag to do condition compiling?

sophia-guo:
Just to confirm that if excluded tests "upgrading an upgradale system/platform module" and CpMpJink test "use of --class-for-name" are excluded for all jdk versions?

openjdk.build/build.xml
Mesbah-Alam: Good catch! It should not be built for Java 8, it can be built for Jaca 9 and up. Need to add a check here. WIll update the PR.

Mesbah-Alam:
Are we still testing Java 9 or 10?  Modularity doesn't exist in Java 8. And I believe it safe to not write and maintain complicated tests for modularity features not supported in Java 11 but only in Java 9 as Modularity had still been evolving during Java 9-- lots of things have been changed since.

openjdk.build/build.xml
Mesbah-Alam: @sophia-guo  - a check has been added.

Mesbah-Alam:
@sophia-guo - agreed. Here's the grinder with JDK 8 - with a modularity test target - it builds successfully and runs nothing : https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/469/console

PR #201, opened 2018-11-28
Fix call to getLocalPart()

Mesbah-Alam:
Fix call to getLocalPart()


Mesbah-Alam:
There were some spaces in the code that were resulting in some special characters being generated while the test is compiled on AIX. This PR fixes it. 

Tested in internal Grinder.

PR #217, opened 2019-01-09
Fix modularity makefile for cygwin

lumpfish:
Fixes https://github.com/eclipse/openj9-systemtest/issues/61

Signed-off-by: Simon Rushton <rushton@uk.ibm.com>

### DIFF
--- a/openjdk.build/build.xml
+++ b/openjdk.build/build.xml
@@ -40,7 +40,7 @@ limitations under the License.
 		<equals arg1="${JDK_VERSION}" arg2="8"/>
 	</condition>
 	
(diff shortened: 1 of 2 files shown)


lumpfish:
The build-modularity target should now be redundant because openjdk.test.modularity/build.xml does its own java version check based on the java.vm.specification.version system property.
So I've removed it, and the dependency on it.  Works OK with java 11 or java 8 on my local machine.

lumpfish:
@Mesbah-Alam  - thanks for testing, merged

Mesbah-Alam:
Tests at Adopt: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/762/console

PR #218, opened 2019-01-11
Disable generation of coredumps at load test failure by default to save space

Mesbah-Alam:
Disable generation of coredumps at load test failure by default to save space

### DIFF
--- a/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/ClassloadingLoadTest.java
+++ b/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/ClassloadingLoadTest.java
@@ -71,6 +71,7 @@ public class ClassloadingLoadTest implements StfPluginInterface {
 				.addPrereqJarToClasspath(JavaProcessDefinition.JarId.JUNIT)
 				.addPrereqJarToClasspath(JavaProcessDefinition.JarId.HAMCREST)
 				.addProjectToClasspath("openjdk.test.classloading")
+				.generateCoreDumpAtFirstLoadTestFailure(false)
 				.setAbortIfOutOfMemory(false)
 				.addSuite("classloading")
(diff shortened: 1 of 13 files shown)


Mesbah-Alam:
Related: https://github.com/eclipse/openj9-systemtest/pull/69#event-2068842129

Mesbah-Alam:
Related: https://github.com/eclipse/openj9/issues/4251

Mesbah-Alam:
Background: After speaking with Peter Shipton, decided to turn off core file generation of these tests. The cores/failures don't cause the tests to fail, so nobody would look at them. Moreover, these core files continue to make builds fail on AIX due to lack of disk space issue.

Mesbah-Alam:
Same change made in multiple tests, tested with Grinder for one test: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/791/console

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/ClassloadingLoadTest.java
```diff
@@ -71,6 +71,7 @@ public void execute(StfCoreExtension test) throws StfException {
 				.addPrereqJarToClasspath(JavaProcessDefinition.JarId.JUNIT)
 				.addPrereqJarToClasspath(JavaProcessDefinition.JarId.HAMCREST)
 				.addProjectToClasspath("openjdk.test.classloading")
+				.generateCoreDumpAtFirstLoadTestFailure(false)
```
sophia-guo: Just confirm if the default value is true? Thanks.

Mesbah-Alam: Thanks for pointing this out and reminding me! By default it is set to false. So we actually do not need this PR. Closing it without delivery.

PR #220, opened 2019-01-29
Add space after -o option in JlinkTest makefile

Mesbah-Alam:
Add space after -o option in JlinkTest makefile
Signed-off-by: Mesbah_Alam@ca.ibm.com <Mesbah_Alam@ca.ibm.com>

PR #221, opened 2019-01-30
Revert "Add space after -o option in JlinkTest makefile"

smlambert:
Reverts AdoptOpenJDK/openjdk-systemtest#220 
It appears to be the cause of native tests failing to compile on Windows, https://github.com/eclipse/openj9-systemtest/issues/81).


smlambert:
Expediting this without review to unblock testing.

PR #244, opened 2019-04-05
Update ASM 5.2 to 7.0 to support JDK11+

adamfarley:
The HCRLateAttachWorkload test won't run on JDK11 if we're not using
ASM 7.0 or up.

This change updates the version of ASM from 5.2 to 7, and also removes
the 6-beta option, as 7 supersedes both.

### DIFF
--- a/openjdk.build/include/top.xml
+++ b/openjdk.build/include/top.xml
@@ -141,11 +141,8 @@ limitations under the License.
 	<path id="tools.class.path">
 		<pathelement location="${prereqs_root}/tools/tools.jar"/>
 	</path>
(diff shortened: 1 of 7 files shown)


adamfarley:
Tested on 11 for OpenJ9 and Hotspot, and also 8.

Will re-test after commit before reintroducing this test into the test buckets for 11+

Mesbah-Alam:
Was this change tested with JDK8? Just want to ensure the new ASM is compatible with JDK8.

adamfarley:
Yes @Mesbah-Alam it was tested with jdk8. See my comment above.

Issue #246, opened 2019-04-08
JlmRemote* failures


jdekonin:
I can reproduce 2 different scenarios with the first test (I didn't look into the others).  One where LT1 appears to complete a workload and exiting cleanly but this causes CL1 to end with an `UndeclaredThrowableException`.  Two where CL1 terminates with very little meanful output to determine cause.

```
03:57:46  	TestJlmRemoteClassNoAuth_0
03:57:46  	TestJlmRemoteMemoryAuth_0
03:57:46  	TestJlmRemoteMemoryNoAuth_0
03:57:46  	TestIBMJlmRemoteClassAuth_0
03:57:46  	TestIBMJlmRemoteClassNoAuth_0
03:57:46  	TestIBMJlmRemoteMemoryAuth_0
03:57:46  	TestIBMJlmRemoteMemoryNoAuth_0
```

Could be related to #145 and #208.  Although the platform I see it on is x86 linux, variant openj9.

Mesbah-Alam:
For scenario 1, what we need to try is increase the workload on LT1 so that it does not end before CL1 finishes. Apparently the workload LT1 is running is not long-running enough, especially in the internal machines.

jdekonin:
This is almost identical to what I am seeing.
https://ci.adoptopenjdk.net/view/Test_system/job/openjdk12_hs_systemtest_ppc64le_linux/23/consoleFull

Mesbah-Alam:
The PPC64LE failure above is different: 
```
CL2 stderr j> 2019/04/08 10:22:08.197   Memory Pool:              G1 Old Gen
CL2 stderr j> 2019/04/08 10:22:08.198   Memory Type:              HEAP
CL2 stderr j> 2019/04/08 10:22:08.199   Peak Usage:               init = 118489088(115712K) used = 19607136(19147K) committed = 120586240(117760K) max = 2147483648(2097152K)
CL2 stderr j> 2019/04/08 10:22:08.199   Current Usage:            init = 118489088(115712K) used = 19874816(19409K) committed = 45088768(44032K) max = 2147483648(2097152K)
CL2 stderr Exception in thread "main" java.lang.AssertionError: Peak Usage used memory smaller than Current Usage used memory
CL2 stderr 	at org.junit.Assert.fail(Assert.java:88)
[...]

PR #247, opened 2019-04-08
Increase workload on server process in JLM tests

Mesbah-Alam:
Increase workload on server process in JLM tests


Mesbah-Alam:
JLM tests that involve a server process running a worklaod while a client process tries to communicate to it via JLM API are sporadically failing across platforms due to the server process finishing too early while the client is still trying to read. 

This PR increases the workload in the server for all JLM tests. It doesn't hurt since once the client finishes, the server process is killed by the test. 

Related: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/246

PR #252, opened 2019-04-12
Increase timeout of subprocess in patch module image test

Mesbah-Alam:
Increase timeout of subprocess in patch module image test


Mesbah-Alam:
The timeout set in these sub-processes is quite short however and times out on AIX. I've increased them from 1m to 5m and ran a 4x grinder in internal Jenkins, and the tests are passing. 

Related : https://github.com/eclipse/openj9/issues/5462

Issue #271, opened 2019-06-10
Mauve issue related to SecurityManager


pshipton:
Not the problem is with tests that enable the SecurityManager, not with the tests that fail.
See also #263

https://ci.eclipse.org/openj9/job/Test_openjdk11_j9_sanity.system_ppc64le_linux_Nightly/59/
```
LT  00:13:54.456 - First failure detected by thread: load-2. Not creating dumps as no dump generation is requested for this load test
LT  00:13:54.461 - Test failed
LT    Failure num.  = 1
LT    Test number   = 3050
LT    Test details  = 'Mauve[gnu.testlet.java.net.URLConnection.getRequestProperties]'
LT    Suite number  = 0
LT    Thread number = 2
LT  >>> Captured test output >>>
LT  Test failed:
LT  java.lang.ExceptionInInitializerError
LT  	at java.base/java.lang.J9VMInternals.ensureError(J9VMInternals.java:193)
LT  	at java.base/java.lang.J9VMInternals.recordInitializationFailure(J9VMInternals.java:182)
LT  	at java.base/sun.net.www.protocol.http.Handler.openConnection(Handler.java:62)
LT  	at java.base/sun.net.www.protocol.http.Handler.openConnection(Handler.java:57)
LT  	at java.base/java.net.URL.openConnection(URL.java:1051)
LT  	at gnu.testlet.java.net.URLConnection.getRequestProperties.test(getRequestProperties.java:43)
LT  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
LT  	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
LT  	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
LT  	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
LT  	at net.adoptopenjdk.loadTest.adaptors.MauveAdaptor.executeTest(MauveAdaptor.java:74)
LT  	at net.adoptopenjdk.loadTest.LoadTestRunner$2.run(LoadTestRunner.java:182)
LT  	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #276, opened 2019-06-11
Exclude sub-tests that use SecurityManager

Mesbah-Alam:
Related https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/271

### DIFF
--- a/openjdk.test.load/config/inventories/mauve/mauve_multiThread_exclude.xml
+++ b/openjdk.test.load/config/inventories/mauve/mauve_multiThread_exclude.xml
@@ -1,4 +1,15 @@
 <inventory>
+	<!-- 
+	Tests that use SecurityManager fail intermittently when run in a 
+	multi-threaded environment due to timing issues. 
+	Issue reference: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/271
+	 -->
+	<mauve class="gnu.testlet.java.security.ProtectionDomain.Implies"/>
+	<mauve class="gnu.testlet.java.net.URLConnection.getRequestProperties"/>
+	<mauve class="gnu.testlet.java.net.URLConnection.getPermission"/>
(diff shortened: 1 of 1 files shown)


sophia-guo:
I'm not familiar with system tests framework. Looks like mauve_all.xml (test on one thread, could also comment out non-working testcase), mauvu_all_exclude.xml( exclude problematic testcase). Same to multi-thread tests. Should those tests be also excluded in one thread tests?

openjdk.test.load/config/inventories/mauve/mauve_multiThread_exclude.xml
```diff
@@ -1,4 +1,15 @@
 <inventory>
+	<!-- 
+	Tests that use SecurityManager fail intermittently when run in a 
+	multi-threaded environment due to timing issues. 
+	Issue reference: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/271
+	 -->
+	<mauve class="gnu.testlet.java.security.ProtectionDomain.Implies"/>
+	<mauve class="gnu.testlet.java.net.URLConnection.getRequestProperties"/>
+	<mauve class="gnu.testlet.java.net.URLConnection.getPermission"/>
+	<mauve class="gnu.testlet.java.net.URLStreamHandler.Except"/>
```
sophia-guo: Should tests that make a call to url.openConnection() from running in Multi-threaded Mauve load be excluded? @pshipton

Mesbah-Alam:
> I'm not familiar with system tests framework. Looks like mauve_all.xml (test on one thread, could also comment out non-working testcase), mauvu_all_exclude.xml( exclude problematic testcase). Same to multi-thread tests. Should those tests be also excluded in one thread tests?

No. The reason why this exclusion is happening only occurs in a multi-threaded scenario; as plained by Pete here: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/271#issuecomment-500621760

Issue #259, opened 2019-05-15
MathLoadTest_bigdecimal_0 incompatible with Java 13


pshipton:
I ran the system tests on an OpenJ9 Java 13 build, and found the following failure. It's likely related to changes in Java 13, and not specific to OpenJ9.

https://ci.eclipse.org/openj9/job/Test_openjdknext_j9_sanity.system_x86-64_windows_OpenJDK/1
```
MLT testFailure: testItem_0740(net.adoptopenjdk.test.bigdecimal.TestSuite060): expected:<-3[5953862697246314162905484746340871359614113505168999319783495360631452156005707752117911726553375634308091790702876492846864265377892836553693509340707503397209982115310256415249098018077865788815173701691026788460916647380644589633161711866424669654959565240828944633747635436183859976250080805236824971673600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000]> but was:<-3[.59538626972463141629054847463408713596141135051689993197834953606314521560057077521179117265533756343080917907028764928468642653778928365536935093407075033972099821153102564152490980180778657888151737016910267884609166473806445896331617118664246696549595652408289446337476354361838599762500808052368249716736E+616]>
MLT org.junit.ComparisonFailure: expected:<-3[5953862697246314162905484746340871359614113505168999319783495360631452156005707752117911726553375634308091790702876492846864265377892836553693509340707503397209982115310256415249098018077865788815173701691026788460916647380644589633161711866424669654959565240828944633747635436183859976250080805236824971673600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000]> but was:<-3[.59538626972463141629054847463408713596141135051689993197834953606314521560057077521179117265533756343080917907028764928468642653778928365536935093407075033972099821153102564152490980180778657888151737016910267884609166473806445896331617118664246696549595652408289446337476354361838599762500808052368249716736E+616]>
MLT 	at org.junit.Assert.assertEquals(Assert.java:115)
MLT 	at org.junit.Assert.assertEquals(Assert.java:144)
[...]

PR #277, opened 2019-07-03
Update BigDecimal Tests for JDK13

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/259


Mesbah-Alam:
Background: 

There has been made a deliberate change in JDK13 in movePointRight():

JDK 13:
https://github.com/ibmruntimes/openj9-openjdk-jdk13/blob/0bcd1d5bd66ca25fba18a7bbf5003a395470d4fe/src/java.base/share/classes/java/math/BigDecimal.java#L2903-L2909

JDK 12:
https://github.com/ibmruntimes/openj9-openjdk-jdk12/blob/b26359aebe7ec913c0abdac75fb6abe21ed28cf0/src/java.base/share/classes/java/math/BigDecimal.java#L2901-L2905

https://bugs.openjdk.java.net/browse/JDK-8183912

This PR updates the tests that use .movePointRight(0) to return the same string as passed in when we are using JDK13 and up.

Mesbah-Alam:
movePointLeft() was also updated in JDK13. So, I'll update the tests that use that API as well.

JDK13:
https://github.com/ibmruntimes/openj9-openjdk-jdk13/blob/0bcd1d5bd66ca25fba18a7bbf5003a395470d4fe/src/java.base/share/classes/java/math/BigDecimal.java#L2879

JDK12:
https://github.com/ibmruntimes/openj9-openjdk-jdk12/blob/b26359aebe7ec913c0abdac75fb6abe21ed28cf0/src/java.base/share/classes/java/math/BigDecimal.java#L2879

llxia:
Is it possible to use existing API to get java version?
https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/89ed48fb3ee763b6559d7802648c8431429d6d25/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/ClassloadingLoadTest.java#L50

Mesbah-Alam:
@llxia - I thought about using the existing API to get the Java version. But the reason why I didn't do that is because, then we'd have to pass in an instance of `StfCoreExtensionc` into these Junit test classes who don't have any reference to StfCoreExtension by default, and who are not invoked directly from any class who instantiate StfCoreExtension either. So, It seemed more work than necessary to use existing APIs and I created a simple class to get the vm spec version instead.

llxia:
Thanks @Mesbah-Alam for testing against openj9 and hotspot in Grinder.

PR #281, opened 2019-08-08
Add check to only build jck project in jck builds 

Mesbah-Alam:


Mesbah-Alam:
We should only be building openjdk.test.jck project when we are trying to run "jck" tests. Otherwise, it will be built any time we are building "systemtest", and the openjdk.test.jck project will fail to build at the native build step (due to all properties not being set - properties that only get set when we run "jck" builds). This PR adds a condition to avoid building openjdk.test.jck when we are running non-jck builds.

Mesbah-Alam:
Tested with "systemtest" build list here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder-systemtest/182/console

PR #282, opened 2019-08-16
Fix relative path for jck excludes

Mesbah-Alam:
Fix relative path for jck excludes


openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
```diff
@@ -207,11 +207,11 @@ public void pluginInit(StfCoreExtension test) throws Exception {
 
 		// Look for an update to the initial excludes file
 		if (jckVersion.contains("jck6") || jckVersion.contains("jck7")) {
```
sophia-guo: In this case there seems we can just remove this if condition?

sophia-guo: Never mind, my bad. There is a character difference.

Mesbah-Alam:
Testing link shared with Sophia on Slack.

PR #290, opened 2019-09-09
Restrict use of tnameserv to only jck8

Mesbah-Alam:
- Related to internal backlog issue # 149
- Restricts use of tnameserver in jck wrapper only within jck 8. tnameserver is a CORBA tool, it has been removed by http://openjdk.java.net/jeps/320

### DIFF
--- a/openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
+++ b/openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
@@ -355,10 +355,13 @@ public class Jck implements StfPluginInterface {
 						.addArg("-J-Dsun.rmi.activation.execPolicy=none " + "-J-Djava.security.policy=" + test.env().findPrereqFile(testSuiteFolder + "/lib/jck.policy").toString())
 						);
 				
-				tnameserv = test.doRunBackgroundProcess("Starting tnameserver", "TNAM", ECHO_ON, ExpectedOutcome.neverCompletes(), test.createJDKToolProcessDefinition()
+				// tnameserv has been removed from jdk11. We only should need it for jck8
+				if (jckVersion.contains("jck8")) { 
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Tested in internal Grinder: 
- with JDK 8, TARGET= sanity.jck : /job/Grinder_Advanced/1112/
- with JDK 11, TARGET=jck-runtime-vm_0: job/Grinder_Advanced/1114/console

sophia-guo:
Is the related issue link correct? I can't find related issue.

Mesbah-Alam:
@sophia-guo - since it's an internal issue, I simply posted the number. I've shared the issue's actual link with you on Slack.

openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
```diff
@@ -663,7 +672,9 @@ public void jckConfigurationForMultijvmWithNoAgent(StfCoreExtension test) throws
 				extraJvmOptions += " -Djdk.xml.maxXMLNameLimit=4000";
 			}	
 			if ( tests.contains("api/org_omg") || tests.contains("api/javax_management") || tests.equals("api") ) {
-				fileContent += "set jck.env.runtime.idl.orbHost " + hostname + ";\n";
+				if (jckVersion.contains("jck8")) {
```
sophia-guo: If this is limited to jck8 probably moving this if condition out of if condition ` if ( tests.contains("api/org_omg") ...`. So it would be same style as following code?

Or maybe also combine those three if jck8 condition to one?

Mesbah-Alam: Updated! Thanks

sophia-guo:
@Mesbah-Alam maybe update comment without using the number sign and number. Otherwise it will be automatic referred as the repo itself's issue link, which could be confusing|misleading to other people.

Mesbah-Alam:
> maybe update comment without using the number sign and number. 
 
Makes sense. Updated!

PR #291, opened 2019-09-10
Add support for BSD/amd64

battleblow:
* Recognise bsd_x86-64 as a valid platform
* Set up make variables to work on bsd_x86-64
  * FreeBSD and OpenBSD have both switched to Clang as default and don't have gcc
  * NetBSD, I believe, still has gcc as default, but 'cc' will still work
  * The include file path jni_md.h uses the full OS name, e.g. 'freebsd', rather than just 'bsd'.
  * Other setup mirrors Linux for the most part

Note that other architectures are possible here (e.g., bsd_x86-32), but are not included in this change.

### DIFF
--- a/openjdk.test.modularity/src/tests/com.test.jlink/native/makefile
+++ b/openjdk.test.modularity/src/tests/com.test.jlink/native/makefile
@@ -52,7 +52,7 @@ P=:
 ###
 # Check platform is set to a single valid value
 ###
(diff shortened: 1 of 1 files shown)


battleblow:
Note that I additionally see errors due to the use of raw & in openjdk.build/playlist.xml.  That shouldn't only affect FreeBSD though, so I haven't added it to this change.  I'll ask about that on the Slack channel to get some broader input.

PR #305, opened 2019-10-27
Replace if (ant-contrib) with if:set (built-in)

aahlenst:
If you don't have Ant-Contrib on your system (it's not listed in the README as a requirement) and because Ant-Contrib is not included in the repository, `openjdk-systemtest-clone-make.sh` won't work because it doesn't know about `<if>` which is provided by Ant-Contrib. I've replaced it with the built-in `if:set` (https://ant.apache.org/manual/ifunless.html).

### DIFF
--- a/openjdk.build/build.xml
+++ b/openjdk.build/build.xml
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
This is fixed in https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/372

PR #312, opened 2019-11-26
Increase threadsTimeout for transformer threads to finish

Mesbah-Alam:
- Increase threadsTimeout for transformer threads to finish - to resolve for intermittent failures. 


Mesbah-Alam:
Test: Test_grinder/job/Grinder/4015/console

llxia:
Instead of increasing the timeout, I would like some further investigation to know why this test failed internally.

llxia:
We do not see a significate machine difference. The plan is to increase the timeout and monitor the build for a while to see if we can collect more information.

PR #320, opened 2020-01-15
Update asm to 7.3

Mesbah-Alam:
- Update ASM references for 7.3 
- Update openjdk.test.debugging build file to add extra options (--enable-preview, -source)  when compiling on jdk11, and up (note JDK 8 doesn't support --enable-preview)
- Related to : https://github.com/eclipse/openj9/issues/8085

### DIFF
--- a/openjdk.test.debugging/build.xml
+++ b/openjdk.test.debugging/build.xml
@@ -46,9 +46,13 @@ limitations under the License.
 	<path id="project.class.path">
 		<path refid="junit.class.path" />
 		<path refid="stf.class.path" />
-		<path refid="asm.7.1.class.path" />
+		<path refid="asm.7.3.class.path" />
 		<path refid="tools.class.path" />
 	</path>
+	
+	<condition property="isJavas8" value="true">
+		<equals arg1="${java.specification.version}" arg2="1.8"/>
+	</condition>
 
 	<!-- Projects which need to be built before this one. -->
 	<!-- dir must be set on the ant task otherwise the basedir property is not set to a new value in the subant task. -->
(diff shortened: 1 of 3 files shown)


Mesbah-Alam:
Test with JDK 8 (where --enable-preview option is not needed for HCRLateAttachWorkload test) : Test_grinder/job/Grinder/5403/console

Test with JDK 8 (Where changes related to addition --enable-preview is in effect) : Test_grinder/job/Grinder/5401/console

openjdk.test.debugging/build.xml
```diff
@@ -113,6 +145,9 @@ limitations under the License.
 			   encoding="${src-encoding}"
 			   includeantruntime="false"
 			   failonerror="true">
+			<compilerarg value="-source"/>
+			<compilerarg value="${java.specification.version}"/>
+			<compilerarg value="--enable-preview"/>
```
smlambert: Is this argument ignored in the jdk11 case (as its in the build-java-11-and-up target)?

Mesbah-Alam: Are you spotting anything in the logic that may prevent  `build-java-11-and-up` from running on JDK 11?. I tested it with JDK 8 and JDK 14 only, did not test it with JDK 11. `build-java-11-and-up` is meant only be executed when we are not running JDK 8 (i.e., running JDK 11 and up - i.e., only the JDKs that support `--enable-preview` option).

smlambert: I just understood --enable-preview to essentially exist to enable preview features in tip (which will be ever-changing, jdk14... then jdk15, etc).  It looked like its being added as an arg to everything jdk11 and up, so was wondering what effect it had in jdk11, which is not tip... and there would be no features to preview.  As long as its benign, we are good to merge.

PR #321, opened 2020-01-23
Use java_platform instead of PLATFORM in JCK build script

Mesbah-Alam:
After https://github.com/AdoptOpenJDK/TKG/pull/30 - we no longer run TKG before compilation in which the map between SPEC --> PLATFORM used to be run. Hence we do not have `PLATFORM` variable available during compilation any more. This PR updates JCK build script to not depend on `PLATFORM`. 

This PR uses `java_platform` (initialized in in top.xml) in place of `PLATFORM` - which servers the same purpose. We already use it in the compilation of modularity test natives. 

Related: backlog/issues/231


Mesbah-Alam:
Tested: Test_grinder/job/Grinder/5611/console

PR #335, opened 2020-03-18
Adding mixed workloads to run as part of Load_Level_4

psoujany:
1) Created a inventory file called mix-few.xml, which contains below workloads.
bigdecimal
mauve_all
lang
nio
concurrent

2) Created a STF equivalent testcase MixedLoadTest.

3) Added MixedLoadTest to openjdk-tests otherLoadTests playlist.

### DIFF
--- a/openjdk.test.load/config/inventories/mix/mix-more.xml
+++ b/openjdk.test.load/config/inventories/mix/mix-more.xml
@@ -1,0 +1,12 @@
+<!-- 
+  This is a minimal inventory intended for tests which just want to be able
+  to generate a load which can be used for testing other capabilities alongside.
+-->
+
+<inventory>
+	<include inventory="/openjdk.test.load/config/inventories/math/bigdecimal.xml"/>
+	<include inventory="/openjdk.test.load/config/inventories/mauve/mauve_all.xml"/>
+	<include inventory="/openjdk.test.load/config/inventories/lang/lang.xml"/>
+	<include inventory="/openjdk.test.load/config/inventories/nio/nio.xml"/>
+	<include inventory="/openjdk.test.load/config/inventories/concurrent/concurrent.xml"/>
+</inventory>
--- a/openjdk.build/makefile
+++ b/openjdk.build/makefile
@@ -475,7 +475,8 @@ test.TestJlmRemoteClassNoAuth \
 test.TestJlmRemoteMemoryAuth \
 test.TestJlmRemoteMemoryNoAuth \
 test.TestJlmRemoteNotifierProxyAuth \
-test.TestJlmRemoteThreadNoAuth 
+test.TestJlmRemoteThreadNoAuth  \
+test.MixedLoadTest
 
 # Tests which won't run on Java 8
 NOT_JAVA8_TESTS:=\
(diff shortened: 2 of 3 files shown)


psoujany:
Successful test run
https://hyc-runtimes-team-svt-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/268/console

openjdk.test.load/config/inventories/mix/mix-more.xml
```diff
@@ -0,0 +1,12 @@
+<!-- 
+  This is a minimal inventory intended for tests which just want to be able
+  to generate a load which can be used for testing other capabilities alongside.
+-->
+
+<inventory>
```
Mesbah-Alam: Could you please explain which existing Load_Level_4 test this STF load replaces?

Mesbah-Alam:
This single test takes `1h41m` to complete. Is this how long the JLTF based Load_Level_4 test that this one replaces used to take as well?

psoujany:
Load_Level_4 contains mix of workloads running with more thread count and multiple times and with timeout 3hours.

Yes, this test replaces 75% of test coverage of JLTF Load_Level_4. Remaining coverage will be done in JLTF itself as part of Load_Level_4 playlist with mix.3hrs test.

Mesbah-Alam:
> Yes, this test replaces 75% of test coverage of JLTF Load_Level_4.

Just for records, please explain "75%" - exactly which tests from the LL4 playlist does this single STF test replace?

psoujany:
Load_Level_4 JLTF has below mixed workloads.
io
nio
mauve
abbs
jni
bigdecimal
harmony
xml
concurrent

STF will be covering
nio(both nio and io)
mauve
lang(abbs)
bigdecimal
concurrent

JLTF will be covering remaining tests
harmony
jni
xml
abbs(few benchmark tests)

Mesbah-Alam:
> STF will be covering
nio(both nio and io)


Looking at [nio.xml ](https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/master/openjdk.test.load/config/inventories/nio/nio.xml), it seems to only list nio tests. Could you please explain where the "io" tests are coming from?

psoujany:
IO workload of JLTF contains below testcases.
testcases.nio.NioBuffersTest
testcases.nio.NioApp
testcases.nio.nioBuffers
testcases.nio.WordCount

These are covered under nio inventory. So I included nio.xml.

Mesbah-Alam:
Thanks, @psoujany 

Non-IBM testing: I've launched a grinder using OpenJ9 JDK 11 on Linux, just to ensure test compilation works fine on non-IBM sdks : SVT-jenkins/Test_grinder/job/Grinder/272/console

smlambert:
Thanks @Mesbah-Alam, it is good as a matter of habit to always check both other open impls, and no reason not to run those Grinder checks at the open server.

Mesbah-Alam:
@psoujany - as Shelley mentioned above, for testing out "open" PRs, we could also use the [open grinder at Adopt ](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/)

just an FYI.

psoujany:
@smlambert 
Could you please provide me access to submit grinders at open server https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/

smlambert:
Yep thanks all!

Adding this as additional notes on verifying PRs, in this case, I would expect at least 3 Grinders:
- 1 internal to check that the new targets work as expected
- 2 external to check that the new targets do not break the 2 main impls

If there were special handling of certain versions or platforms in playlists, I may add extra Grinders to check that they worked on those jdk versions and platforms.

Mesbah-Alam:
Thanks Shelley.. 

I had missed running a test on Hotspot! Here it is : svt-jenkinns/Grinder/273/console

The compilation is fine on hotspot, so we are good to go.

PR #338, opened 2020-04-01
jck11 requires -source 11 instead of -source 9

JasonFengJ9:
**jck11 requires -source 11 instead of -source 9**

This is expected to fix following error.
```
warning: as of release 10, 'var' is a restricted local variable type and cannot be used for type declarations or as the element type of an array
        SAM s = (var i) -> i;
                 ^
../tests/lang/LMBD/lmbd182/lmbd18201m001/lmbd18201m001.java:21: error: cannot find symbol
        SAM s = (var i) -> i;
                 ^
  symbol:   class var
Number of errors: 1
```
Just added `jck11` case and didn't modified existing `jck9` just in case it is need for whatever reason. Assuming `jck10` is outdated.

Reviewer: @smlambert 


smlambert:
Thanks @JasonFengJ9 !

PR #349, opened 2020-06-06
Detect and tolerate execution on RISC-V

sxa:
Have tested this and it allows the `_MachineInfo` test to run :-)

Signed-off-by: Stewart Addison <sxa@uk.ibm.com>

### DIFF
--- a/openjdk.test.modularity/src/tests/com.test.jlink/native/makefile
+++ b/openjdk.test.modularity/src/tests/com.test.jlink/native/makefile
@@ -52,7 +52,7 @@ P=:
 ###
 # Check platform is set to a single valid value
 ###
-VALID_PLATFORMS?=osx_x86-64,aix_ppc-32,aix_ppc-64,linux_390-32,linux_390-64,linux_ppc-32,linux_ppc-64,linux_x86-32,linux_x86-64,win_x86-32,win_x86-64,zos_390-32,zos_390-64,linux_ppcle-64,linux_arm-32,linux_arm-64,bsd_x86-64
+VALID_PLATFORMS?=osx_x86-64,aix_ppc-32,aix_ppc-64,linux_390-32,linux_390-64,linux_ppc-32,linux_ppc-64,linux_x86-32,linux_x86-64,win_x86-32,win_x86-64,zos_390-32,zos_390-64,linux_ppcle-64,linux_arm-32,linux_arm-64,bsd_x86-64,linux_riscv64-64
 ifndef PLATFORM
   $(error "The variable PLATFORM needs to be defined")
 endif


smlambert:
If I am not mistaken, there were a couple of other places that need updates ... but maybe that got corrected finally (to avoid having to change more than one spot).

sxa:
Yep Shelley I'm creating PRs for other repos at the moment, but this is at least enough to let the MachineInfo target to work so is at least a valid first step :-)

smlambert:
Cool, if you search for BSD in this repo, STF and in openj9-systemtest you will find all the spots:
https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/291
https://github.com/AdoptOpenJDK/stf/pull/67
etc

sxa:
I suspect s390x might be a better example as BSD picks up all sorts of references that are nothing to do with FreeBSD suppot ;-)

sxa:
(Note that this excludes the JCK makefile as we have no way to test those changes that at adopt just now ...)

sxa:
Not sure my brain's working too well this weekend but for some reason https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3273/parameters/ didn't work (with this PR) and https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3274/parameters/ (without this PR)

I think I should stop and come back to this on Tuesday since I'm making silly mistakes somewhere.

smlambert:
I have not looked at your Grinder, but you have several changes across the different test repos that likely need to be tested together which is tricky but do-able (the compile job is failing to find contents of openj9-systemtest repo in the location it expects, despite git clone of the master branch appears to work).  

I think we can merge the openjdk-tests PR first, which will make co-ordinating the rest of the PRs slightly easier.

Mesbah-Alam:
> https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3273/parameters/ didn't work (with this PR) and https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3274/parameters/ (without this PR)

In https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3273/parameters/  - `STF_REPO` is not pointing to your fork, but AdoptOpenJDK instead, while you are setting a developer branch in `STF_BRANCH`. Please set `STF_REPO=https://github.com/sxa/stf`. 

Also, please note that there exists another makefile with a similar list of supported platforms : https://github.com/eclipse/openj9-systemtest/blob/00eb7d2f0e1bfb35872072edf9db56b53223d8c7/openj9.test.sharedClasses.jvmti/src/native/Makefile#L69

Most likely you will likely have to add `linux_riscv64-64` here too.

sxa:
Pretty sure this can be canceled now as we had it working elsewhere, so I'll close this

PR #359, opened 2020-08-04
Exclude SolverTest

adamfarley:
This test fails as per the referenced issue. Excluding the
test for now to make triage faster.


Mesbah-Alam:
We had decided in the past to not add excludes of sub-tests from a load test run that is specific to a particular Java Implementation or version; as this process ends up in creation of many exclude files and checks in test code itself. The decision was : if we do need to exclude a sub-test, we should simply exclude it from everywhere. I believe this is what we should do here as well. 

FYI @lumpfish

adamfarley:
@Mesbah-Alam - Change made.

Mesbah-Alam:
Grinder worked fine: [InternalJenkins]/Grinder/9466

PR #362, opened 2020-09-24
Added JCK16 config files

Mesbah-Alam:
Added JCK16 config files


Mesbah-Alam:
@JasonFengJ9 - please review.

openjdk.test.jck/config/jck16/runtime.jti
```diff
@@ -0,0 +1,187 @@
...
+jck.env.runtime.awt.crossWindowFocusTransferSupported=Yes
+jck.env.runtime.awt.robotAvailable=Yes
+jck.env.runtime.awt.splashScreenOpt=-splash\:\#
+jck.env.runtime.awt.splashScreenSupported=Yes
+jck.env.runtime.awt=Yes
+jck.env.runtime.fp.doubleMaxExponent=
+jck.env.runtime.fp.doubleMinExponent=
+jck.env.runtime.fp.floatMaxExponent=
+jck.env.runtime.fp.floatMinExponent=
+jck.env.runtime.fp.format=Intel, 16 bit exponent
```
JasonFengJ9, line 38: This should be `jck.env.runtime.fp.format=Intel, 15 bit exponent` for all JCK levels.

Mesbah-Alam: Updated. Please check. (I'll make the same update in other JCK levels using a separate PR).

JasonFengJ9: Others are good. Maybe `15` was search/replaced by `16` along with other `JCK15` specific changes.

JasonFengJ9 (approved):
LGTM

llxia:
It looks to me the config files are very similar. And most of the changes are either comments or descriptions.
```
# TESTSUITE example: /jck/jck16/JCK-compiler-16
```
```
jck.env.description=JCK16 compiler template jti for STF automation
```

What is the use of these files? Can we have a general file, so we do not have to update this for every java version?

Mesbah-Alam:
> What is the use of these files? 

These template JTI files are generated (once per JCK level) by running the JCK GUI and manually answering the "interview questions" therein. The STF JCK wrapper uses these files to auto-generate the "JTB" files-- which contain a similar, but more specific (e.g. contains current machine information) list of key=value pairs, applicable for a specific JCK run. 

Some JTI files can be more different than others. For example, the JTI files for JCK 8 may include [devtools related parameters ](https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/master/openjdk.test.jck/config/jck8c/compiler.jti#L66,) which will be absent in JTI files of later JCK levels. 

>  Can we have a general file, so we do not have to update this for every java version?

This will certainly be a nice to have. To do this, we'd have to update STF JCK wrapper (Jck.java) to programmatically handle each JCK version and what unique set of parameters is required by it, so that the right JTB file gets generated. 

Perhaps this enhancement could be planned together with the work we have in mind to remove the STF JCK wrapper?

llxia:
> Some JTI files can be more different than others. For example, the JTI files for JCK 8 may include devtools related parameters which will be absent in JTI files of later JCK levels.

Do we have lots of these special cases? If not, could we have a set of default files? If there is a special case, create files for that java version (i.e., jck8c). Otherwise, use the default files. Does this make sense?

I think the change just needs to happen at the following. 
https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/2d9430d2bfa085d0074d8ce23b22a6d46c5da1ca/openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java#L210

And yes, ultimately, we want to remove the STF JCK wrapper. Before we get there, we should do simple improvements to make our life easier.

Mesbah-Alam:
Sounds good. I'll have a look.

llxia:
Thanks @Mesbah-Alam . We have the similar logic for [openjdk tests](https://github.com/AdoptOpenJDK/openjdk-tests/blob/6bf96a0d8462edd9a8e4442a3d99da1c9a8958b2/openjdk/openjdk.mk#L147-L150). Th only difference is that it uses make.

Mesbah-Alam:
@llxia , We have specific properties that vary across quite a number of JCK versions. For example, in case of runtiume.jti,  `jck.tests.setOfModules` lists a unique set of modules which is different in 9, 11, 13,.. etc. We also have properties that are only available in certain JCK levels, for example `jck.env.testPlatform.testsOnPreviewFeaturesEnabled` is avaiable from 12+. 

So, if we want to have a default template for runtime.jti for example, we'd have to address these special cases programmatically; e.g., we'd have to add logic in Jck.java that will read in the default runtime.jti values from the template file, and if jckVersion=12+, add the extra parameter and spit out a generated runtime.jti file, which will then be passed onto the JCK command line to be executed.

Again, for `jck.tests.setOfModules` we need to add logic to put the correct set of modules as the value of `jck.tests.setOfModules` in the generated runtime.jti file.. etc .

The same process needs to be repeated for compiler.jti and devtools.jti too. 

**** 

So, given the above, I am wondering which option is better:  

a) Keeping separate JTI files for each Jck version  
    - Pro: No extra logic in Jck.java - simply pass the static jti files to JCK command line
    - Con: need to add one set of JTI files whenever we add new JCK version 
b)  Use a common template
     - Pro: Don't need to add a new set of JTI file each time we add a new JCK version 
     - Con: We'd have to update Jck.java to add logic to handle special cases, likely every time we introduce new JCK versions

llxia:
In general, when we have a default template, we use the one that is the most common. From what I can see, 12-16 are very similar.

Mesbah-Alam:
"From what I can see, 12-16 are very similar." - 

Yes, they are similar, but not identically - they have some parameters with unique values, as explained [above](https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/362#issuecomment-700801221). So, if we use the template, we'd have to programmatically add those unique values before we can use them.

llxia:
Do we have the unique values in 12 -16? Assuming `# TESTSUITE example` or `jck.env.description` does not matter.

PR #363, opened 2020-10-01
Add null check to StringTransformer

mikezhang1234567890:
If a class is defined after an instance of StringTransformer is registered,
the transform method will be called and passed a null className as the new
class is being defined. This will throw an exception and end the test so we
need to ensure the className is not null before the equality check.

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>

Fixes https://github.com/eclipse/openj9/issues/10504


Mesbah-Alam (approved):
LGTM

Mesbah-Alam:
Thanks @mikezhang1234567890 . The change looks good.

PR #364, opened 2020-10-08
Fix jlmlocal and jlmthread tests

lumpfish:
Fixes https://github.com/eclipse/openj9/issues/1566

### DIFF
--- a/openjdk.test.jlm/src/test.jlm/net/adoptopenjdk/test/jlm/resources/ThreadData.java
+++ b/openjdk.test.jlm/src/test.jlm/net/adoptopenjdk/test/jlm/resources/ThreadData.java
@@ -448,6 +448,7 @@ public class ThreadData extends VMData {
 			CompositeData[] lockThdDataset = null;
 			CompositeData[] dumpThdDataset = null;
 
+			Report.printlnIndent("Invoking invokeMBeanOperation: 1");
 			dumpThdDataset = (CompositeData[])invokeMBeanOperation(mbs, tb, "dumpAllThreads", 
 					new Object[] { monitorSupported, synchSupported });
 
@@ -793,6 +794,7 @@ public class ThreadData extends VMData {
 						Report.decrementIndentation();
 					}
 					Report.decrementIndentation();
+					Report.printlnIndent("Invoking invokeMBeanOperation: 2");
 					invokeMBeanOperation(mbsc, objectName, mbci.getName(), arguments);
 					ExpectedMBeanInfo.checkOffConstructor(mbci.getName());
 				} // End of "if(empty mbpia) ... else ..."
@@ -883,6 +885,7 @@ public class ThreadData extends VMData {
 						Report.incrementIndentation();
 						Report.printlnIndent("returns \"" + mboi.getReturnType() + "\"");
 						Report.decrementIndentation();
+						Report.printlnIndent("Invoking invokeMBeanOperation: 3");
 						invokeMBeanOperation(mbsc, objectName, mboi.getName(), arguments);
 					}
 					ExpectedMBeanInfo.checkOffOperation(mboi.getName());
@@ -930,6 +933,7 @@ public class ThreadData extends VMData {
 		if (argType.equals("int")) {
 			return new Integer(1);
 		}
+		// Assume a long parameter type is expecting a thread id
 		if (argType.equals("long")) {
 			return new Long(Thread.currentThread().getId());
 		}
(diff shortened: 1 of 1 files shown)


openjdk.test.jlm/src/test.jlm/net/adoptopenjdk/test/jlm/resources/ThreadData.java
```diff
@@ -1010,68 +1022,95 @@ public Object invokeMBeanOperation(MBeanServerConnection mbsc, ObjectName object
...
+									Report.printlnSuccessMsg("MBean operation " + operationName + " not invoked - filtered out in ThreadData.java");
+								}
+								if (methodSignature != null
+									&& !operationName.equals("getThreadInfo")
+									&& !operationName.equals("getNativeThreadIds") //This requires "monitor" permission - which is not configured
+									&& ( !operationName.equals("dumpAllThreads") && arguments.length == 3 ) // Known openj9 defect. Remove check when https://github.com/eclipse/openj9/issues/10796 is fixed
+									&& supported == true) {
+									Report.printlnSuccessMsg("Attempting to invoke MBean operation " + operationName);
+									boolean parameterTypesMatch = parameterTypesMatch(mbpia, arguments);
+//									if ( parameterTypesMatch(mbpia, arguments) ) {
```
Mesbah-Alam, line 1088: Please remove the commented out line here.

openjdk.test.jlm/src/test.jlm/net/adoptopenjdk/test/jlm/resources/ThreadData.java
```diff
@@ -1094,16 +1133,72 @@ public Object invokeMBeanOperation(MBeanServerConnection mbsc, ObjectName object
 			if (rmbe.getCause() instanceof IllegalArgumentException) {
 				Report.printlnInformationMsg(
 						"Invocation failed: IllegalArgumentException (probably guessed arguments badly)");
+				rmbe.printStackTrace();
```
Mesbah-Alam, line 1136: Should we fail the test in this case as well?

lumpfish: Are you saying you don't think we should fail?  I can't see how we will ever know the condition is being hit unless the test fails.  The test pipelines would 'pass' so the detailed test logs (where the message is written) would never be examined.

The test thinks it has generated suitable arguments for the mxbean operation being invoked - so if that is not the case either something has changed (the argument values are no longer suitable) or there is a new mxbean operation / signature which the test is not (yet) aware of, so failing should be a good way to alert us to any such changes.

Mesbah-Alam: I was wondering if we should add an Assert.fail(..) over there. Currently it's only printing the stack trace, but not failing the test, in that particular case.

lumpfish: Ah - sorry, I assumed your comment applied to the code I had changed (where I added an Assert.fail() (line 1104).

I did not change th e code you mention - but since the RuntimeMBeanException is now caught in a try - catch block specific to the `invoke` method (line 1094) I believe the code in the `catch RuntimeMBeanException` block on line 1131 is now unreachable and should have been removed.  Please check if you agree - if so. I'll update the PR.

openjdk.test.jlm/src/test.jlm/net/adoptopenjdk/test/jlm/resources/ThreadData.java
```diff
@@ -1010,68 +1022,95 @@ public Object invokeMBeanOperation(MBeanServerConnection mbsc, ObjectName object
...
+								}
+								if (methodSignature != null
+									&& !operationName.equals("getThreadInfo")
+									&& !operationName.equals("getNativeThreadIds") //This requires "monitor" permission - which is not configured
+									&& ( !operationName.equals("dumpAllThreads") && arguments.length == 3 ) // Known openj9 defect. Remove check when https://github.com/eclipse/openj9/issues/10796 is fixed
+									&& supported == true) {
+									Report.printlnSuccessMsg("Attempting to invoke MBean operation " + operationName);
+									boolean parameterTypesMatch = parameterTypesMatch(mbpia, arguments);
+//									if ( parameterTypesMatch(mbpia, arguments) ) {
+									try {
```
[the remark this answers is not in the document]
lumpfish: > Is the code inside this try block (line 1090) the only code that may generate a RuntimeMBeanException? If yes we can get rid of the catch block for RuntimeMBeanException further down (line 1131).

Yes.
I have removed the catch block at line 1131 and confirmed the tests still run OK: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4185/

>  In that case, we should add an Assert.fail() inside the `if ( ! parameterTypesMatch )` block after line 1099.

I have deliberately called all overloaded methods even though the arguments we are supplying may be incorrect.  If the arguments are correct then the operation should succeed.  If the arguments are incorrect then `IllegalArgumentException` should be thrown.  The `parameterTypesMatch` check is there to enable the correct behavior to be determined.  So there is no need to fail the test if the `IllegalArgumentException` is thrown and that is what should happen.

Mesbah-Alam:
Approving assuming the commented out line (https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/364#discussion_r501892980) will be removed before merging.

Mesbah-Alam (approved):
LGTM

lumpfish:
The commented out line highlighted in (#364 (comment)) now removed.

PR #384, opened 2020-12-10
Add a time-based STF system test

Mesbah-Alam:
- Adds a time-based STF system test

Related to https://github.com/AdoptOpenJDK/openjdk-tests/issues/2104

FYI @lumpfish 

### DIFF
--- a/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
+++ b/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
@@ -22,15 +22,34 @@ import net.adoptopenjdk.stf.processes.ExpectedOutcome;
 import net.adoptopenjdk.stf.processes.definitions.JavaProcessDefinition;
 import net.adoptopenjdk.stf.processes.definitions.LoadTestProcessDefinition;
 import net.adoptopenjdk.stf.runner.modes.HelpTextGenerator;
+import net.adoptopenjdk.stf.util.TimeParser;
 import net.adoptopenjdk.loadTest.InventoryData;
+import net.adoptopenjdk.stf.environment.StfTestArguments; 
 
 public class MixedLoadTest implements StfPluginInterface {
+	
+	String timeLimit;
+	int testCountMultiplier;
+	boolean isTimeBasedLoadTest = true;
+	String defaultTimeout = "3h";
+	
 	public void help(HelpTextGenerator help) throws StfException {
 		help.outputSection("MixedLoadTest runs workloads of math,mauve,nio,lang,concurrent unit tests");
 		help.outputText("");
 	}
 
 	public void pluginInit(StfCoreExtension stf) throws StfException {
+		// When used as a time based test, this STF test plugin accepts the following two parameters:
+		//   (1) timeLimit     	     : Time duration (e.g. 5s, 5m, 5h) for which the load should be run. 
+		//   (2) testCountMultiplier : Optional multiplier value to increase the workload externally.
+		StfTestArguments testArgs = stf.env().getTestProperties("timeLimit=[notset]","testCountMultiplier=[1]");
+		timeLimit = testArgs.get("timeLimit");
+		testCountMultiplier = Integer.parseInt(testArgs.get("testCountMultiplier"));
+		
+		// When We are using MixedLoadTest for default cases;. e.g. not in a time based load test scenario 
+		if ( timeLimit.equals("notset")) {
+			isTimeBasedLoadTest = false; 
+		}
 	}
 
 	public void setUp(StfCoreExtension test) throws StfException {
@@ -48,9 +67,7 @@ public class MixedLoadTest implements StfPluginInterface {
 		if (!fileSystemJar.asJavaFile().exists()) {
 			throw new StfException("The filesystem.jar does not exist: " + fileSystemJar.getSpec());
 		}
-		
-		String timeout = "3h";
-		
+
 		String inventoryFile = "/openjdk.test.load/config/inventories/mix/mix-more.xml";
 		int totalTests = InventoryData.getNumberOfTests(test, inventoryFile);
 		int cpuCount = Runtime.getRuntime().availableProcessors();
@@ -63,16 +80,37 @@ public class MixedLoadTest implements StfPluginInterface {
 				.addProjectToClasspath("openjdk.test.nio")
 				.addProjectToClasspath("openjdk.test.concurrent")
 				.addJarToClasspath(mauveJar)
-				.addJarToClasspath(fileSystemJar)
-				.setAbortIfOutOfMemory(false)
-				.addSuite("mix")
-				.setSuiteThreadCount(cpuCount - 2, 10)	
-				.setSuiteNumTests(totalTests * 200) // Leave 1 cpu for the JIT. i for GC and set min 20			
-				.setSuiteInventory(inventoryFile) 			// Point at the file which lists the tests
-				.setSuiteRandomSelection();		  			// Randomly pick the next test each time
+				.addJarToClasspath(fileSystemJar);
 				
+		if (isTimeBasedLoadTest) { 
+			loadTestInvocation = loadTestInvocation.setTimeLimit(timeLimit);	// If it's a time based test, stop execution after given time duration
+		}
+		loadTestInvocation = loadTestInvocation.setAbortIfOutOfMemory(false)
+				.addSuite("mix")
+				.setSuiteThreadCount(cpuCount - 2, 10); // Leave 1 cpu for the JIT, 1 cpu for GC and set min 10
+		if (isTimeBasedLoadTest) { 
+			loadTestInvocation = loadTestInvocation
+				.setSuiteNumTests(totalTests * 80000000 * testCountMultiplier); // If it's a time based test, run a very large load
+		} else {  
+			loadTestInvocation = loadTestInvocation
+				.setSuiteNumTests(totalTests * 200);
+		}
+		
+		loadTestInvocation = loadTestInvocation
+				.setSuiteInventory(inventoryFile) 		// Point at the file which lists the tests
+				.setSuiteRandomSelection(); 			// Randomly pick the next test each time
+		
+		String finalTimeout = defaultTimeout;
+		
+		// If it's a time-based test, set timeout for the test process to the default timeout(3h) + given timeLimit
+		if (isTimeBasedLoadTest) { 
+			long finalTimeoutInSeconds = TimeParser.parseTimeSpecification(defaultTimeout).getSeconds() + 
+					TimeParser.parseTimeSpecification(timeLimit).getSeconds();
+			finalTimeout = finalTimeoutInSeconds + "s"; 
+		}
+		
 		test.doRunForegroundProcess("Run mixed unit tests", "LT", Echo.ECHO_ON,
-				ExpectedOutcome.cleanRun().within(timeout), 
+				ExpectedOutcome.cleanRun().within(finalTimeout), 
 				loadTestInvocation);		
 			
 	}
--- a/openjdk.build/makefile
+++ b/openjdk.build/makefile
@@ -811,7 +811,7 @@ test.TestJlmRemoteThreadNoAuth:
 	echo Target $@ completed	
 test.MixedLoadTest:
 	echo Running target $@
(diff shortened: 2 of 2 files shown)


Mesbah-Alam:
Test : InternalGrinder/12632

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/TimedLoadTest.java
```diff
@@ -0,0 +1,92 @@
...
+import net.adoptopenjdk.stf.processes.definitions.LoadTestProcessDefinition;
+import net.adoptopenjdk.stf.runner.modes.HelpTextGenerator;
+import net.adoptopenjdk.loadTest.InventoryData;
+import net.adoptopenjdk.stf.environment.StfTestArguments; 
+
+public class TimedLoadTest implements StfPluginInterface {
+	String timelimitInMins;
+	int testCountMultiplier;
+	public void help(HelpTextGenerator help) throws StfException {
+		help.outputSection("MixedLoadTest runs workloads of math,mauve,nio,lang,concurrent unit tests");
```
llxia, line 32: Is this `TimedLoadTest ` or `MixedLoadTest `?

Mesbah-Alam: Good catch! That was a copy paste error. I've updated the help message.

lumpfish:
I do not think 'TimedLoadTest' is what this test should be called - assuming you might want to create other workload tests which are timed based what would you call the next one?

As far as I can see, this is actually just a variant of MixedLoadTest running for a period of time rather rather than a fixed number of tests.  So why not just modify MixedLoadTest to accept the timeLimit argument?  If the argument is not received, then the test would behave as is does now; if the argument is present it would be used as the timeLimit and the number of tests to be run increased so that they do not complete before the timelimit expires.

Also, why only allow the user to specify the timeLimit in minutes?  The STF -timeLimit option actually uses a 5s / 5m / 5h syntax so it makes sense for the -test-args argument to use the same syntax.

If that were done then a playlist or test makefile could have several targets for the same workload such as:
```
MixedLoadTest:
	stf.pl -test=MixedLoadTest
MixedLoadTest_5secs:
	stf.pl -test=MixedLoadTest -test-args="timeLimit=5s"
MixedLoadTest_5mins:
	stf.pl -test=MixedLoadTest -test-args="timeLimit=5m"
MixedLoadTest_5hrs:
	stf.pl -test=MixedLoadTest -test-args="timeLimit=5h"
```

lumpfish (changes requested):
See comment in PR Conversation regarding requested changes.

Mesbah-Alam:
@lumpfish - Thanks for the review comments. The following changes have been made based on them: 
1) Removed TimedLoadTest altogether. 
2) Updated MixedLoadTest to handle the case where the input parameters are passed in - in which case it will set the time cap, otherwise the default test load will be run
3) Updated the [playlsit](https://github.com/AdoptOpenJDK/openjdk-tests/pull/2107/files) to pass in value like `5m` as accepted by STF -timeLimit option. 

Tested time-bsased case here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5483/ 
Tested default case here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5486/console

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
```diff
@@ -23,14 +23,31 @@
...
 		help.outputSection("MixedLoadTest runs workloads of math,mauve,nio,lang,concurrent unit tests");
 		help.outputText("");
 	}
 
 	public void pluginInit(StfCoreExtension stf) throws StfException {
+		// When used as a time based test, this STF test plugin accepts the following two parameters:
+		//   (1) timelimitInMins     : Time duration (e.g. 5s, 5m, 5h) for which the load should be run. 
+		//   (2) testCountMultiplier : Optional multiplier value to increase the workload externally.
+		StfTestArguments testArgs = stf.env().getTestProperties("timelimitInMins=[0m]","testCountMultiplier=[1]");
+		timelimitInMins = testArgs.get("timelimitInMins");
```
lumpfish, line 44: Since the user can now spcify any time units (5ms, 5s, 5m or 5h) the variable should not have the 'InMins' ending.

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
```diff
@@ -63,13 +80,25 @@ public void execute(StfCoreExtension test) throws StfException {
...
+			loadTestInvocation = loadTestInvocation
+				.setSuiteNumTests(totalTests * 80000000 * testCountMultiplier); 	// If it's a time based test, run a very large load
+		} else {  
+			loadTestInvocation = loadTestInvocation
+				.setSuiteNumTests(totalTests * 200);	// Leave 1 cpu for the JIT. i for GC and set min 20
+		}
+		
+		loadTestInvocation = loadTestInvocation
+				.setSuiteInventory(inventoryFile) 	// Point at the file which lists the tests
+				.setSuiteRandomSelection(); 		// Randomly pick the next test each time
```
lumpfish, line 101: The only values different for a time limited test are the time limit and the number of tests.  It would be easier to follow if only `setTimeLimit` and `setSuiteNumTests` were outside the initlal `loadTestInvocation` definition.

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
```diff
@@ -63,13 +80,25 @@ public void execute(StfCoreExtension test) throws StfException {
...
 				
 		test.doRunForegroundProcess("Run mixed unit tests", "LT", Echo.ECHO_ON,
 				ExpectedOutcome.cleanRun().within(timeout),
```
lumpfish, line 104: `timeout` is hard coded to `3h`.  It should be adjusted for a time limited test.  The time limit is when stf stops submitting new tests, so they will need time to complete.  To achieve roughly the same as currently behavior `timeout` should be `timeout + timelimit`.

lumpfish (changes requested):

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
Mesbah-Alam: I had tried that, but there was a runtime error from STF:  

```
GEN stderr Exception in thread "main" net.adoptopenjdk.stf.StfException: Java invocation built out of sequence. LOAD_TEST_ARGS cannot be set after SUITE
```

That's why the sequence was kept as it is.

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
Mesbah-Alam: Makes sense. Updated the PR.

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
Mesbah-Alam: Is this test ever going to be used to run a load that's 3 hour long (or longer)? The original requirement was to write a STF load test that runs for 5 minutes. That's why the original `3h` timeout was left as is. 

If we plan to run really large loads such as 3h, we need to first make sure that the workload set would last that long.  (the current workload here was only tested to last 20 minutes. It may last much longer-- but I haven't tested it past the 20 minute mark). 

That said, I've updated the test to use a timeout value of `default timeout (3h) + timeLimit`

Mesbah-Alam:
Latest changes were tested here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5489/

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
```diff
@@ -22,15 +22,34 @@
...
 	public void pluginInit(StfCoreExtension stf) throws StfException {
+		// When used as a time based test, this STF test plugin accepts the following two parameters:
+		//   (1) timeLimit     	     : Time duration (e.g. 5s, 5m, 5h) for which the load should be run. 
+		//   (2) testCountMultiplier : Optional multiplier value to increase the workload externally.
+		StfTestArguments testArgs = stf.env().getTestProperties("timeLimit=[0m]","testCountMultiplier=[1]");
+		timeLimit = testArgs.get("timeLimit");
+		testCountMultiplier = Integer.parseInt(testArgs.get("testCountMultiplier"));
+		
+		// When We are using MixedLoadTest for default cases;. e.g. not in a time based load test scenario 
+		if ( timeLimit.equals("0m") && testCountMultiplier == 1) {
```
lumpfish, line 50: Since you are using the default value to test whether timlimit was specified, the default value could be more descriptive (`unspecified` or `notset` or something?).

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
```diff
@@ -63,16 +80,37 @@ public void execute(StfCoreExtension test) throws StfException {
...
+		}
+		loadTestInvocation = loadTestInvocation.setAbortIfOutOfMemory(false)
+				.addSuite("mix")
+				.setSuiteThreadCount(cpuCount - 2, 10); 
+		if (isTimeBasedLoadTest) { 
+			loadTestInvocation = loadTestInvocation
+				.setSuiteNumTests(totalTests * 80000000 * testCountMultiplier); // If it's a time based test, run a very large load
+		} else {  
+			loadTestInvocation = loadTestInvocation
+				.setSuiteNumTests(totalTests * 200);	// Leave 1 cpu for the JIT. i for GC and set min 20
```
lumpfish, line 96: Comment is wrong (looks like it was wrong before this PR - it should be on the `.setSuiteThreadCount(cpuCount - 2, 10)` line, and should say `min 2` not `min 20`.

lumpfish (changes requested):

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
Mesbah-Alam: Done. Test: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5490/console

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
Mesbah-Alam: Yes, this comment was wrong from before. Thanks for catching! Updated.

lumpfish (approved):
Looks good.  Please "Squash and Merge" when you're ready.

openjdk.test.load/src/test.load/net/adoptopenjdk/stf/MixedLoadTest.java
```diff
@@ -63,16 +80,37 @@ public void execute(StfCoreExtension test) throws StfException {
...
 				
+		if (isTimeBasedLoadTest) { 
+			loadTestInvocation = loadTestInvocation.setTimeLimit(timeLimit);	// If it's a time based test, stop execution after given time duration
+		}
+		loadTestInvocation = loadTestInvocation.setAbortIfOutOfMemory(false)
+				.addSuite("mix")
+				.setSuiteThreadCount(cpuCount - 2, 10); // Leave 1 cpu for the JIT, 1 cpu for GC and set min 10
+		if (isTimeBasedLoadTest) { 
+			loadTestInvocation = loadTestInvocation
+				.setSuiteNumTests(totalTests * 80000000 * testCountMultiplier); // If it's a time based test, run a very large load
```
fjeremic, line 93: @Mesbah-Alam shouldn't we just set this to `INT_MAX` or `LONG_MAX` or whatever the maximum representable integer is rather than using a "very large load" with a magic 80,000,000 constant? The tests will be terminated after the alloted time anyway.

Mesbah-Alam: Thanks for your suggestion. I will try this out.

Issue #389, opened 2021-01-06
MathLoadTest_all_special doesn't timeout and collect diagnostics


pshipton:
https://ci.eclipse.org/openj9/job/Test_openjdk8_j9_special.system_ppc64_aix_Nightly_testList_4/34
MathLoadTest_all_special_17
The test appears to hang, but STF doesn't timeout and collect any diagnostics, instead running for hours.
```
23:44:16  MLT 23:44:16.525 - Completed 9.7%. Number of tests started=507 (+55)
23:44:39  MLT 23:44:36.580 - Completed 10.0%. Number of tests started=523 (+16)
23:44:58  MLT 23:44:56.567 - Completed 12.5%. Number of tests started=658 (+135)
23:45:15  STF 23:45:14.023 - Heartbeat: Process MLT is still running
23:45:17  MLT 23:45:16.587 - Completed 15.0%. Number of tests started=787 (+129)
23:50:16  STF 23:50:14.149 - Heartbeat: Process MLT is still running
23:55:26  STF 23:55:14.208 - Heartbeat: Process MLT is still running
00:00:23  STF 00:00:14.296 - Heartbeat: Process MLT is still running
00:05:18  STF 00:05:14.400 - Heartbeat: Process MLT is still running
00:10:29  STF 00:10:14.453 - Heartbeat: Process MLT is still running
00:15:24  STF 00:15:14.498 - Heartbeat: Process MLT is still running
00:20:20  STF 00:20:14.034 - Heartbeat: Process MLT is still running
00:25:28  STF 00:25:14.089 - Heartbeat: Process MLT is still running
00:30:21  STF 00:30:14.149 - Heartbeat: Process MLT is still running
00:35:15  STF 00:35:14.189 - Heartbeat: Process MLT is still running
00:40:26  STF 00:40:14.267 - Heartbeat: Process MLT is still running
00:45:28  STF 00:45:14.343 - Heartbeat: Process MLT is still running
00:50:29  STF 00:50:14.495 - Heartbeat: Process MLT is still running
00:55:29  STF 00:55:14.097 - Heartbeat: Process MLT is still running
01:00:28  STF 01:00:14.229 - Heartbeat: Process MLT is still running
01:05:18  STF 01:05:14.370 - Heartbeat: Process MLT is still running
01:10:18  STF 01:10:14.473 - Heartbeat: Process MLT is still running
[...]

PR #391, opened 2021-01-14
Set timeout of sub-test to a reasonable value

Mesbah-Alam:
This PR reduces the 'math' variant of math load test's timeout value from `20h` to `2h`- which is a more practical timeout value. This prevents cases like the one reported [here](https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/389#issuecomment-756862024) where the test took more than 5 hours to complete. 

Related to https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/389


Mesbah-Alam:
Test: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5946/console

Mesbah-Alam:
Note: We will be adding a [time-based equivalent ](https://github.com/AdoptOpenJDK/openjdk-tests/issues/2104#issuecomment-756926880) of these load tests which should cover the testing that was potentially being covered by the amount of workload which is taken off by this PR.

smlambert (approved):
thanks @Mesbah-Alam !

PR #393, opened 2021-01-15
Have a reduced workload in MathLoadTest and ClassLoadinngLoadTest for all slow running JIT options

Mesbah-Alam:
- The logic of how the tests figure out when slow running JIT modes are in effect has been modified. Now the test will treat any JIT option that contains `-Xjit:count` or `Xjit:enableOSR` (e.g. Mode614, Modew553, etc and the OSR modes) as slow running options and use a reduced workload. This will now ensure we do not run the "regular" (longer) load when any of these JIT modes are in use.

- Related to https://github.com/eclipse/openj9/issues/11666

### DIFF
--- a/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/ClassloadingLoadTest.java
+++ b/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/ClassloadingLoadTest.java
@@ -43,22 +43,25 @@ public class ClassloadingLoadTest implements StfPluginInterface {
 	}
 
 	public void pluginInit(StfCoreExtension stf) throws StfException {
-		if (stf.isJavaArgPresent(Stage.EXECUTE, "-Xjit:count=0,optlevel=warm,gcOnResolve,rtResolve") ||
-			stf.isJavaArgPresent(Stage.EXECUTE, "-Xjit:enableOSR,enableOSROnGuardFailure,count=1,disableAsyncCompilation")) {
-				specialTest = true;
-			}
-			
-			if(specialTest) {
-				// If we are running with special JIT modes which are slow, reduce workload 
-				testCountMultiplier = 1500; 
-				
-			} 	
 	}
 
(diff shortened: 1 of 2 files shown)


Mesbah-Alam:
Test: internal Grinder/13201

Mesbah-Alam:
@peter_shipton@ca.ibm.com - please review.

pshipton:
Not sure this is a desirable change. It seems this will affect modes which already don't run long enough to find JIT bugs, and cause them to run for an even shorter time. I suggest we just target the problematic long running tests until we get the time based testing going.

Mesbah-Alam:
> I suggest we just target the problematic long running tests until we get the time based testing going.

The way to do that would be to hard-code each special JIT option used in each mode inside the test so that it knows that it should run the lighter workload when those modes are in effect.

pshipton:
I think that's the best solution until the time based testing is running. Otherwise the usefulness of testing the other modes is reduced.

pshipton:
> hard-code each special JIT option

atm we're only talking about fixing Mode614 so it's not a big list.

pshipton:
The faster we can roll out time based testing, the sooner we can be done with futzing with the number of iterations.

Mesbah-Alam:
Latest updates to the PR are tested here: Internal Grinder/13217

Mesbah-Alam:
@pshipton : I've updated the PR. : 
1) The workload numbers have been set to what they were originally set to. 
2) The logic of how the tests figure out when slow running JIT modes are in effect has been modified. Now the test will treat any JIT option that contains `-Xjit:count` or `Xjit:enableOSR` (e.g. Mode614, Modew553, etc and the OSR modes) as slow running options and use a reduced workload. This will now ensure we do not run the "regular" (longer) load when any of these JIT modes are in use. Previously the test was only looking for a limited set of JIT options and consequently failing to treat modes like 614 as slow running ones. 

Latest test : Internal Grinder/13224

pshipton (approved):

PR #395, opened 2021-01-19
Increase individual test timeout from 10 min to 20 min

JasonFengJ9:
Each test's timeout limit is multiplied by the time factor value. One base time limit is 10 minutes, increase it to 20 minutes.
This is reflected by the `timeoutSeconds` value within `jtr` file.

Note: there were individual tests take slightly more than 10 min to finish in certain `ppc64le_linux` machines. Verified with this PR, the test in question passes in 10x grinder.


JasonFengJ9:
@Mesbah-Alam could you please review?

Mesbah-Alam (approved):
LGTM

Mesbah-Alam:
@JasonFengJ9 : Approved. However, if this is not a fix for a defect, please defer merging this until after the release that is happening this week.

PR #430, opened 2021-06-17
Use Sun Studio to compile on Solaris

gdams:
fixes: https://github.com/adoptium/aqa-systemtest/issues/429

### DIFF
--- a/openjdk.test.jck/src/native/makefile
+++ b/openjdk.test.jck/src/native/makefile
@@ -140,8 +140,20 @@ SRC_PATH=$(SRCDIR)/src
 # From jck10 the unix include files are in src/share/lib/jni/include/linux
 
 CC=gcc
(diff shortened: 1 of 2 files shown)


openjdk.test.jck/src/native/makefile
```diff
@@ -140,7 +140,7 @@ SRC_PATH=$(SRCDIR)/src
 # From jck10 the unix include files are in src/share/lib/jni/include/linux
 
 CC=gcc
-CFLAGS=-fPIC -I$(SOLARIS_PATH) -I$(LINUX_PATH) -I$(SRC_PATH)
+CFLAGS=-fPIC -I$(LINUX_PATH) -I$(SRC_PATH)
```
karianna, line 143: Why Linux path and not solaris path?

gdams: This is the default CFLAGS for Linux (I moved the Solaris block to it's own IF statement)

llxia (approved):
LGTM

PR #431, opened 2021-06-23
force natives to be compiled in 64bit on solaris

gdams:
Currently, the native libs on Solaris are being compiled in 32bit rather than 64bit. This PR add's the `-m64` flag to force the compiler to use 64bit.

fixes:

```bash
wrong ELF class: ELFCLASS32 (Possible cause: architecture word width mismatch)
```


sxa (approved):
Approving on the basis this has now run through ok on one of the machines.

PR #432, opened 2021-06-29
Add UNIX_PATH to default CFLAGS

Mesbah-Alam:
For JCK8 to compile, we need to pass into the CFLAGS the path to the folder containing `jckjni_md.h`.  This was removed by https://github.com/adoptium/aqa-systemtest/pull/430/files#diff-ce706353a6b9b131babe2b82433446abac32dc9cfd44b16cabd4c5272d7724dbL143. 

In this PR, to remove confusion, we are using the `UNIX_PATH` variable in the default CFLAG values to supply the path to the headers - which solves the JCK8 compile issue. For later versions we do not see the issue as for those we have the "linux" folder which contains the headers. 

### DIFF
--- a/openjdk.test.jck/src/native/makefile
+++ b/openjdk.test.jck/src/native/makefile
@@ -140,7 +140,7 @@ SRC_PATH=$(SRCDIR)/src
 # From jck10 the unix include files are in src/share/lib/jni/include/linux
 
 CC=gcc
-CFLAGS=-fPIC -I$(LINUX_PATH) -I$(SRC_PATH)
+CFLAGS=-fPIC -I$(UNIX_PATH) -I$(LINUX_PATH) -I$(SRC_PATH)
 LDFLAGS=-shared 
 
 ifeq ($(OS),osx)


Mesbah-Alam:
Test : Grinder/16656/

Mesbah-Alam:
@JasonFengJ9 - please review.

openjdk.test.jck/src/native/makefile
```diff
@@ -140,7 +140,7 @@ SRC_PATH=$(SRCDIR)/src
 # From jck10 the unix include files are in src/share/lib/jni/include/linux
 
 CC=gcc
-CFLAGS=-fPIC -I$(LINUX_PATH) -I$(SRC_PATH)
+CFLAGS=-fPIC -I$(UNIX_PATH) -I$(LINUX_PATH) -I$(SRC_PATH)
```
JasonFengJ9, line 143: The `unix` folder within `UNIX_PATH=$(JNI_INCLUDE_PATH)$(D)unix` was introduced as JCK patch, not in default JCK materials.
`JCK16+` has `linux` folder now, can `JCK8/11` change `unix` to `linux` hence only `LINUX_PATH` is required?

Mesbah-Alam: Sure, that's another way of doing it. I will create two PRs - one for 8 and one for 11 to update unix to linux.

Mesbah-Alam: Hi Jason, turns out that your suggestion is harder to test. Because updating jck8 repo with the folder change, somehow is not getting picked up in the git update (re: Grinder/16661)- meaning.. if we make the update in the jck8 and jck11 repos with folder name change, we’d have to potentially remove JCK8 and JCK11 root folders / local repo contents from every machine attached to Jenkins, for the folder change update to be picked up in fresh checkout of those repos.

We need to investigate why JCK update logic is not picking up the folder change. But that investigation can happen a bit later after we deliver the change to decouple JCK from STF. In the interim, to unblocks JCK8 and JCK11 testing, it’s easiest to deliver this PR.

PR #433, opened 2021-06-29
Re-introduce SOLARIS_PATH for linux TCK compilation

gdams:
`jckjni_md.h` is actually located in the Solaris directory but is needed on Linux as well

I removed this as part of https://github.com/adoptium/aqa-systemtest/pull/430


smlambert (approved):
thx @gdams !

PR #440, opened 2021-07-15
Use custom security.properties for SSL TLS1.0,1.1

andrew-m-leonard:
jck api/javax_net tests TLS1.0 & 1.1 so must be enabled to support those tests.

Signed-off-by: Andrew Leonard <anleonar@redhat.com>

### DIFF
--- a/openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
+++ b/openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
@@ -324,6 +324,15 @@ public class Jck implements StfPluginInterface {
 		
 		test.doEchoFile("The JTB file generated\n", newJtbFileRef);
 
+		if ( tests.contains("api/javax_net") ) {
+			// Requires TLS 1.0/1.1 enabling
+			DirectoryRef secPropsLocation = test.env().getResultsDir().childDirectory("SecProps");
+			test.doMkdir("Creating dir to store the custom security properties", secPropsLocation);
+			FileRef secPropsFileRef = secPropsLocation.childFile("security.properties");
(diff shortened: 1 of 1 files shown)


gdams (approved):

andrew-m-leonard:
I've tested with jdk-11 and 16 as well

openjdk.test.jck/src/test.jck/net/adoptopenjdk/stf/Jck.java
```diff
@@ -324,6 +324,15 @@ public void setUp(StfCoreExtension test) throws Exception {
 		
 		test.doEchoFile("The JTB file generated\n", newJtbFileRef);
 
+		if ( tests.contains("api/javax_net") ) {
```
Mesbah-Alam, line 327: Just wondering, do these changes require any set up to be done somewhere else for them to work?

andrew-m-leonard:
i’ve tested win & mac on jck8, 11 and 16

Mesbah-Alam (approved):
LGTM

Mesbah-Alam:
@andrew-m-leonard - could you please let us know if this fixes all the javax_net issues? 

This target has been disabled on 8 and 11 for ibm+openj9: https://github.com/adoptium/aqa-tests/blob/391642bdca37e09fdca74b0743353383915b337c/jck/runtime.api/playlist.xml#L935 

If this issue fixes all javax_net issues, we should un-exclude the target.

smlambert:
We will not be able to confirm that @Mesbah-Alam, for 2 reasons
1) we do not have visibility of backlog/issues/461 
2) we are not allowed to verify this against non-Temurin builds.  I suggest that you check on an internal grinder by running BUILD_LIST=jck, TARGET=disabled.jck-runtime-api-javax_net now that this is merged

PR #441, opened 2021-07-20
Update AIX parameters in TCK makefile to allow native builds

sxa:
I'd be ok if we decided to switch to using `xlc` from the PATH, but the natives are currently trying to reference a nonexistant `linux` directory. I believe `amd64` will be ok for this as a default directory to use, but we'll see how well the latest test run goes. It could probably use a bit of a subsequent variable name refactor since using `OSX_PATH` on AIX is rather odd ...

We may wish to check this with anyone else utilising this part of the `aqa-systemtest` repository.

Signed-off-by: Stewart X Addison <sxa@redhat.com>

### DIFF
--- a/openjdk.test.jck/src/native/makefile
+++ b/openjdk.test.jck/src/native/makefile
@@ -156,8 +156,8 @@ ifeq ($(OS),sunos)
 endif
 
 ifeq ($(OS),aix)
-	CC=xlc
-	CFLAGS=-qstaticinline -q$(BitMode) -I$(UNIX_PATH) -I$(LINUX_PATH) -I$(SRC_PATH)
+	CC=/opt/IBM/xlC/13.1.3/bin/xlc
+	CFLAGS=-qstaticinline -q$(BitMode) -I$(UNIX_PATH) -I$(OSX_PATH) -I$(SRC_PATH)
 	LDFLAGS=-G -q$(BitMode)
 endif
 


Mesbah-Alam:
@sxa - We usually create PRs for aqa-systemtest repo using branches off of developer forks, not master. Could the same be done for this PR? 

And since it's a change for AIX, a grinder link with at least one system test run - proving the change will not break system test compilation would be great. 

Thanks!

smlambert:
Grinder for this PR:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1130

We won't share link to TCK grinder job as its only accessible to Temurin-compliance committers.  (tagged you in Mesbah, so that you are aware & could check that it doesn't 'break' you internally).

Branched off of master, as sxa's fork is named openjdk-systemtest (which he'll have to delete and recreate at some point in the future to be able to run Grinder tests from).

Mesbah-Alam:
This PR assumes that IBM xlC 13.1.3 be available on all AIX machines. @AdamBrousseau - is this a valid assumption for the internal AIX machines? 

Note: I ran a single grinder internally, and it does work, Grinder/17038. Just want to confirm that it'd work on all other AIX machines.

Mesbah-Alam:
Ran a job to check the internal AIX machines, and they do seem to have IBM xlC 13.1.3 installed. The PR looks good to me.

Mesbah-Alam (approved):
LGTM

sxa:
I may change this back to `xlc` in the future and insist that the compiler is on the `PATH` but this is the safest option for now to ensure we have stability of the configuration on the machines - thanks for confirming it'll work for you @Mesbah-Alam 👍🏻

smlambert (approved):
thanks all!

PR #444, opened 2021-07-30
Change JCK8 AIX to use cmdAsString rather than cmdAsFile

andrew-m-leonard:
This previous fix for JCK11 does not apply to JCK8 as OsHelper class is jck11+ only: https://github.com/adoptium/aqa-systemtest/commit/af676759f9365f29533837b466bb1a38a92be4db

Added condition to only use workaround for jck11+.

Signed-off-by: Andrew Leonard <anleonar@redhat.com>


smlambert (approved):

andrew-m-leonard:
Test successfully on jck8.

PR #445, opened 2021-08-11
Update JCK Natives compile flags to define _AMD64_ on Windows

sxa:
Fixes an issue building the TCK material

Signed-off-by: Stewart X Addison <sxa@redhat.com>


sxa:
> So just to be clear - `_AMD64_` needs to be defined for both 32 bit and 64 bit natives?

That's likely a question for @d3r3kk specifically, but it seems reasonable if it's an architecture definition and I would be extremely surprised if it caused a problem on 64-bit (it's definitely REQUIRED for 32-bit per the slack thread)

sxa:
(To be clear on my previous statement I have tested the COMPILATION in a 64-bit setup and it still builds the dll correctly, but I haven't run any tests against it)

lumpfish (approved):
Looks good

PR #455, opened 2021-08-30
Correct path to stf.pl in build document

keithc-ca:
It's not at the top level, but in the `stf.core/scripts` folder.


llxia (approved):
Thanks @keithc-ca

PR #457, opened 2021-09-16
Add Authorization Code Grant OAuth security test

joeyleeeeeee97:
#### Secure Communication
Test the data that travels across a network can not be accessed by someone who is not the intended recipient, and test varities combination of 
cryptography and application protocol works together.

- User Story
> Assume server is an small application needs to read data from github, and github access requires a token to use its public API.
> Normally, users won't trust the 3rd party application with their account and password, so we need to redirect user to the github login website
and callback to app with logined token.
> But still, token is very important and shouldn't be transported via plain text, so we introduce Authorization Code and let the application backend 
> to request the token with specific app secret.

- But the Authorization Code is still text, what if someone try to get a token via that code?
> OAuth center must verify code and app secret to ensure only the application could get token.

- Is this a common usage?
> Authorization Code Grant 
> https://www.oauth.com/oauth2-servers/server-side-apps/authorization-code/


joeyleeeeeee97:
https://github.com/adoptium/aqa-systemtest/issues/453

tellison:
I'll let the test experts speak to the details of this, but it looks like a great contribution to me! Thank you.

smlambert (commented):
Thanks for this PR @joeyleeeeeee97 !  I am still reviewing this, but thought I would submit some initial comments, which are minor grammar/spelling in comments.  There are also a couple of binary files in this PR.  Do you mean to include the .DS_store file? 

As part of a review, reviewers will want to run some Grinder jobs in Jenkins.  For this and for inclusion into our regular automated testing, there will need to be a complementary PR in the aqa-tests repo.  

Do you plan to create a PR in aqa-tests to hook this test into our AQA automation?  Likely in its own subdirectory called security inside https://github.com/adoptium/aqa-tests/tree/master/system with a build.xml to instruct how to fetch/build this material and playlist.xml for the test target to execute.

joeyleeeeeee97:
@smlambert  Thanks for the suggestions and yes I am working on a PR in aqa-tests to include this test.

Mesbah-Alam (approved):
LGTM

smlambert (approved):
Most October release activities have finished, so we can now merge this and its related aqa-tests PR in, thanks @joeyleeeeeee97 !

PR #458, opened 2021-09-16
Add support for compiling natives on macOS arm64

gdams:
fixes failing `vm_jni` tests on macOS arm64. The `-arch arm64` flag is needed because the shell runs in x86_64 emulation


karianna (approved):
LGTM - can you confirm this still works on Intel-based hosts?

gdams:
> LGTM - can you confirm this still works on Intel-based hosts?

Yeah I've already done that! :D

PR #473, opened 2021-12-17
Exclude gnu.testlet.java.net.http url connection.request properties test

Mesbah-Alam:
- Exclude gnu.testlet.java.net.http url connection.request properties test
- Related to openj9/issues/14168


Mesbah-Alam:
Test : Grinder/20028

Mesbah-Alam:
@Peter_Shipton@ca.ibm.com - please review.

pshipton:
Does it need to be excluded for all versions when it works up to and including jdk17, or can it be excluded for jdk18 and later?

Mesbah-Alam:
> or can it be excluded for jdk18 and later? 

The only way to do that is to have separate load configuration xml files. We went that route in the past and ended up having to maintain a host of different configuration files for different versions. We subsequently decided to stop doing that. 

Also, since these mauve tests are primarily used to stress the JVM as "load tests", our focus has been to have a load that works on all versions.

pshipton (approved):

PR #475, opened 2022-04-22
Exclude nio from zos

Mesbah-Alam:
In order to avoid running into encoding issues that the generated class files of NIO load test pose (see openj9-openjdk-jdk11-zos/issues/342 for details), while on z/OS we will run a mixed load test that does not include Nio load tests. 

This PR creates a new inventory file to be used by MixedLoadTest while on z/OS which essentially uses the same loads as before minus NIO tests. 


Mesbah-Alam:
Test on z/OS :  Grinder/23049 (Passes with the new load on z/OS). 
TEst on Linux: Grinder/23048.

pshipton:
lgtm

PR #478, opened 2022-08-31
Reintroducing StringIndexOutOfBoundsException as the bug is fixed

adamfarley:
This test should now run and pass on all JDK versions it's meant to
run against, as the upstream OpenJDK bug (JDK-8267773) is resolved
and backported, with the code fully propagated.

Signed-off-by: Adam Farley <adfarley@redhat.com>


adamfarley:
Fix verified: https://ci.adoptopenjdk.net/job/Grinder/5550/console

That link proves that the test was reintroduced with this fix, and also verifies that the test now passes.

karianna (approved):
LGTM

smlambert:
If this is reintroducing the test for all implementations, we need a Grinder to verify openj9 impl is also green.

smlambert:
FYI @llxia @Mesbah-Alam

PR #489, opened 2024-09-23
Build the tests with release=8 whenever it is possible

fridrich:
Fixes https://github.com/eclipse-openj9/openj9/issues/19888

The tests have to be built with release=8 in order to avoid calling the new overridden methods with covariant return types in java.nio.ByteBuffer/CharBuffer/LongBuffer, like position(int), rewind, flip...
The only source file that has to be built with source=1.8 and target=1.8 is the gnu/testlet/java/lang/Class/classInfo/getDeclaredMethod.java, because it requires sun.reflect.annotation.AnnotationType that is not visible using release=8 from jdk9+
The specifying of release="8", source="1.8" and target="1.8" allows ant 1.10.x to use either release if supported or the source/target if release is not supported.

### DIFF
--- a/openjdk.test.mauve/build.xml
+++ b/openjdk.test.mauve/build.xml
@@ -106,10 +106,11 @@ limitations under the License.
 			<exclude name="/junit/textui/*.java" />
 		</depend>
 		-->
-		<!-- The test cases are built with source="1.8" and target="1.8" because they need packages
-		    which have been moved into madules which are not visible by default in Java 9. -->
+		<!-- The test cases are built with release="8" if supported by the compiler,
+		    otherwise with source="1.8" and target="1.8". -->
 		<javac srcdir="${openjdk_test_mauve_src_dir}"
 			   destdir="${openjdk_test_mauve_src_dir}"
+			   release="8"
 			   source="1.8"
 			   target="1.8"
 			   fork="true"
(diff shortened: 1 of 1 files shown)


karianna:
@fridrich Are you able to sign the Eclipse CLA?

llxia:
Thanks @fridrich for providing the fix. Could you try to sign ECA? https://github.com/adoptium/aqa-tests/blob/master/Contributing.md#eclipse-contributor-agreement

It should be a straightforward process. Unfortunately, we cannot accept the change without the ECA signed.

llxia:
@annaibm Please help to test this PR. Thanks

pshipton:
FYI the comment which I'll copy here.

[fridrich](https://github.com/fridrich) commented [Sep 23, 2024](https://github.com/eclipse-openj9/openj9/issues/19888#issuecomment-2369060838)

So, that pull request basically fixes the build of the mauve.jar, so that it works with the java 8. It is hanging in eclipsefdn/eca. I am not sure whether we (SUSE) have ever signed something like this. I also don't care to be credited, so anybody who has the right paperwork on file can submit that patch. I transfer automatically the copyright to anybody brave enough to get it integrated.

fridrich:
At the end, I signed it in my personal capacity. So it should be good I think.

llxia:
Quick update: We encountered other issues while upgrading to JDK 17 for compilation for all dependent jars (though `mauve.jar` is fine). This PR can't be merged until those issues are resolved.

smlambert (changes requested):
Given the comment that it can't be merged, putting a request changes 'hold' on this one so that it is not accidentally merged ahead of the other issues are resolved (more details on the other issues referenced would be good to link to this @llxia )

llxia (approved):
LGTM

llxia:
@smlambert we are ready to merge this PR.

smlambert (approved):

PR #490, opened 2024-10-02
Removed build-java-8

annaibm:
- Due to the error in reading tools.jar, as it is not available in newer JDK versions,
- Added tools.jar and jdi.jar for java 8 from lib folder. and it aligns with newer JDKs (JDK 9 and above) which no longer include tools.jar, preventing related build failures.


openjdk.test.debugging/build.xml
```diff
@@ -48,8 +48,12 @@ limitations under the License.
...
 		<path refid="junit.class.path" />
 		<path refid="stf.class.path" />
-		<path refid="asm.class.path" />
-		<path refid="tools.class.path" />
+		<path refid="asm.class.path" />		
+	</path>
+
+	<path id="project.class.path8">
+		<path refid="project.class.path"/>
+		<path refid="tools.class.path" />
```
llxia, line 56: This is the same as before. Please revert.

PR #494, opened 2025-01-29
Exclude Mauve tests that use SecurityManager for JDK24

theresa-m:
The SecurityManager is removed in JDK 24. Exclude Mauve tests that are failing due to illegal use of the security manager.


llxia (approved):
Thanks @theresa-m

LongyuZhang (approved):
LGTM

PR #504, opened 2025-10-03
Remove -b option

smlambert:
Resolve error on nodes where older git versions do not support -b option

```
get-source:
     [exec] unknown option: -b master
     [exec] usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
     [exec]            [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
     [exec]            [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
     [exec]            [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
     [exec]            [--super-prefix=<path>] [--config-env=<name>=<envvar>]
     [exec]            <command> [<args>]
     [exec] Result: 129
```


sophia-guo:
Might need <arg value="clone"/> ?

PR #505, opened 2025-10-03
Fix clone command

smlambert:
Original PR was wrong (missing 'clone'), so all other options were reporting failure.  Noting that getDependency jobs do not have aqa-systemtest parameter, making it less convenient to test, thus the churn).


llxia:
https://ci.adoptium.net/job/systemtest.getDependency/242/

PR #506, opened 2025-10-03
Fix clone

smlambert:


openjdk.test.mauve/build.xml
```diff
@@ -362,8 +362,6 @@ limitations under the License.
 	<target name="get-source" depends="create-work-dir" unless="openjdk_test_mauve_already_built">
 		<exec dir="${openjdk_test_mauve_work_dir}" executable="git">
 		<arg value="clone"/>
-		<arg value="--depth 1"/>
-		<arg value="-b master"/>
```
keithc-ca, line 366: To pass two arguments, you need to use two `<arg/>` elements or use the `line` attribute:
```
		<arg value="--depth"/>
		<arg value="1"/>
```
See https://ant.apache.org/manual/using.html#arg.

PR #513, opened 2026-05-25
Add more debugging for AIOOBE to MessageHeader.parse()

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/17575#issuecomment-4536437630

Related to https://github.com/adoptium/aqa-systemtest/pull/512

Tested in https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/5191/

### DIFF
--- a/openjdk.test.nio/src/test.nio/net/adoptopenjdk/test/nio2/util/MessageHeader.java
+++ b/openjdk.test.nio/src/test.nio/net/adoptopenjdk/test/nio2/util/MessageHeader.java
@@ -37,13 +37,14 @@ public class MessageHeader {
 				// Key value are delimited by ':'
 				String[] keyValue = parts[index].split(":", 2);
 				if (keyValue.length != 2) {
+					System.out.println("Malformed header line - '" + parts[index] + "'");
 					System.err.println("Malformed header line - '" + parts[index] + "'");
 				}
 				try {
 					map.put(keyValue[0], keyValue[1]);
 				} catch (ArrayIndexOutOfBoundsException e) {
 					ArrayIndexOutOfBoundsException boundsex = new ArrayIndexOutOfBoundsException("parts[" + index + "] '" + parts[index] + "' split length "
-							+ keyValue.length + " [0] '" + keyValue[0] + "'");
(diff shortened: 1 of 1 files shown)


openjdk.test.nio/src/test.nio/net/adoptopenjdk/test/nio2/util/MessageHeader.java
keithc-ca: Suggest this message also be printed to `System.out`.

pshipton: added

openjdk.test.nio/src/test.nio/net/adoptopenjdk/test/nio2/util/MessageHeader.java
```diff
@@ -37,13 +37,13 @@ public void parse(String headerString) {
 				// Key value are delimited by ':'
 				String[] keyValue = parts[index].split(":", 2);
 				if (keyValue.length != 2) {
-					System.err.println("Malformed header line - '" + parts[index] + "'");
+					System.out.println("Malformed header line - '" + parts[index] + "'");
```
keithc-ca, line 40: I was suggesting it be written to _both_ stderr and stdout.

pshipton: What is the point, we aren't seeing the output to stderr, and may not see the output to stdout either. We already know the header line is malformed and only has 1 part. Once we see the entire header, included in the Exception message which we do see, we can think about what sort of change we should make to the test, or where there is a problem that needs to be investigated further.

keithc-ca: The point was to ensure that message was together with the stack trace (no matter whether we're looking at stdout or stderr). If it shows up neither place, I think that would support my theory that a JIT problem is at play.

pshipton: updated it.

keithc-ca (approved):

